PR #3, opened 2021-11-02
porting functionality from sortingview and add colors

magland:
porting various functionalities from sortingview
* elec geom for avg waveforms
* sortable table widget for units table
* select units for spike amp view (lockable)

Also put in some color per unit

### DIFF
--- a/src/views/AverageWaveforms/AverageWaveformsView.tsx
+++ b/src/views/AverageWaveforms/AverageWaveformsView.tsx
@@ -4,6 +4,7 @@ import React, { FunctionComponent, useCallback, useMemo } from 'react';
 import { AverageWaveformsViewData } from './AverageWaveformsViewData';
 import AverageWaveformPlot from './AverageWaveformPlot';
 import colorForUnitId from 'views/common/colorForUnitId';
+import {mean} from 'mathjs';
 
 type Props = {
     data: AverageWaveformsViewData
@@ -57,11 +58,6 @@ const subtractChannelMeans = (waveform: number[][]) => {
     })
 }
 
-const computeMean = (x: number[]) => {
-    if (x.length === 0) return 0
-    let sum = 0
-    for (let a of x) sum += a
-    return sum / x.length
-}
+const computeMean = (ary: number[]) => ary.length > 0 ? mean(ary) : 0
 
 export default AverageWaveformsView
--- a/src/views/AverageWaveforms/AverageWaveformsView.tsx
+++ b/src/views/AverageWaveforms/AverageWaveformsView.tsx
@@ -22,7 +22,7 @@ const AverageWaveformsView: FunctionComponent<Props> = ({data, width, height}) =
         label: `Unit ${aw.unitId}`,
         props: {
             channelIds: aw.channelIds,
-            waveform: aw.waveform,
+            waveform: subtractChannelMeans(aw.waveform),
             samplingFrequency: data.samplingFrequency,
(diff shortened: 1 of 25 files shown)


src/views/AverageWaveforms/AverageWaveformsViewData.ts
```diff
@@ -20,13 +21,15 @@ export type AverageWaveformsViewData = {
     averageWaveforms: AverageWaveformData[]
     samplingFrequency: number
     noiseLevel: number
+    channelLocations?: {[key: string]: number[]}
```
jsoules, line 24: Could this be anything other than 2d?

src/views/SpikeAmplitudes/SelectUnitsWidget.tsx
```diff
@@ -0,0 +1,50 @@
+import NiceTable from 'components/NiceTable/NiceTable'
```
jsoules, line 1: This code is all pretty much what was in SortingView, right? (I haven't looked at it too closely, on the assumption that it's the code we had before.)

jsoules (commented):
Couple questions, mostly about the mean subtraction for average waveforms.

src/views/AverageWaveforms/AverageWaveformsView.tsx
```diff
@@ -47,4 +50,18 @@ const AverageWaveformsView: FunctionComponent<Props> = ({data, width, height}) =
...
+        return W.map(a => (a - mean0))
+    })
+}
+
+const computeMean = (x: number[]) => {
+    if (x.length === 0) return 0
+    let sum = 0
+    for (let a of x) sum += a
+    return sum / x.length
+}
```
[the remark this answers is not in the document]
magland: Okay, I'll make that change.

src/views/AverageWaveforms/AverageWaveformsView.tsx
```diff
@@ -17,18 +18,20 @@ const AverageWaveformsView: FunctionComponent<Props> = ({data, width, height}) =
...
     }, [setSelectedUnitIds])
-    const plots = useMemo(() => (data.averageWaveforms.map(aw => ({
+    const plots = useMemo(() => (data.averageWaveforms.sort((a1, a2) => (a1.unitId - a2.unitId)).map(aw => ({
         key: `${aw.unitId}`,
         label: `Unit ${aw.unitId}`,
+        labelColor: colorForUnitId(aw.unitId),
         props: {
             channelIds: aw.channelIds,
-            waveform: aw.waveform,
+            waveform: subtractChannelMeans(aw.waveform),
```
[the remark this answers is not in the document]
magland: Yes, the reason to do this... if the raw traces is not bandpass filtered, then the avg waveform can have an offset far from zero, and for display purposes we'd like to subtract off the mean for each channel. It might not be the best decision, but it was needed for certain datasets I was working with. We can revisit down the road.

src/views/AverageWaveforms/AverageWaveformsViewData.ts
magland: It could be 3d, but we're only displaying 2d.

src/views/SpikeAmplitudes/SelectUnitsWidget.tsx
magland: yes, i pretty much copied this over

PR #5, opened 2021-11-04
implement curation

magland:
Added a sorting curation context
Implemented the curation control in the left panel
sortingCurationUri property on the mountain layout data
Curation actions are dispatched to the backend via action tasks

### DIFF
--- a/src/views/MountainLayout/MountainLayoutView.tsx
+++ b/src/views/MountainLayout/MountainLayoutView.tsx
@@ -1,5 +1,8 @@
+import SortingCurationAction from 'contexts/SortingCurationAction';
+import SortingCurationContext, { sortingCurationReducer } from 'contexts/SortingCurationContext';
+import { initiateTask, useSubfeedReducer } from 'figurl';
 import MountainWorkspace from 'MountainWorkspace/MountainWorkspace';
-import React, { FunctionComponent, useMemo } from 'react';
+import React, { FunctionComponent, useCallback, useMemo } from 'react';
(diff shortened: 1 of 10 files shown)


src/views/MountainLayout/MountainLayoutView.tsx
```diff
@@ -20,14 +23,38 @@ const MountainLayoutView: FunctionComponent<Props> = ({data, width, height}) =>
...
+          functionId: 'spikesortingview.sorting_curation_action.1',
+          kwargs: {
+            sorting_curation_uri: data.sortingCurationUri,
+            action: a
+          },
+          functionType: 'action',
+          onStatusChanged: () => {}
+        })
+        // this might be how we can do offline-first curation (get curationSubfeed from useSubfeedReducerS)
+        // curationSubfeed.appendOfflineMessages([a]) // this would need to be implemented
```
jsoules, line 48: So this would essentially be batching the changes locally and then pushing them all at once? Seems like that could be handy. We'd just need to define what happens if the remote feed status has updated before the batch is pushed--might require intelligent intervention on the user's part, but hopefully rare enough that it won't be too much of a problem.

jsoules (commented):
This all looks good--seems like mostly copying over from what was in the previous SortingView, plus adding the context, yeah?

magland:
> This all looks good--seems like mostly copying over from what was in the previous SortingView, plus adding the context, yeah?

Yes that's right.

PR #6, opened 2021-11-04
PositionPlot

magland:
Implement PositionPlotView
Plot animal position over time

builds on curation PR

### DIFF
--- a/src/views/PositionPlot/PositionPlotView.tsx
+++ b/src/views/PositionPlot/PositionPlotView.tsx
@@ -1,0 +1,147 @@
+import { matrix, multiply } from 'mathjs'
+import React, { FunctionComponent, useCallback, useMemo } from 'react'
+import colorForUnitId from 'views/common/colorForUnitId'
+import TimeScrollView, { computePanelDimensions, computePixelsPerSecond, get1dTimeToPixelMatrix, TimeScrollViewPanel } from '../RasterPlot/TimeScrollView/TimeScrollView'
+import { PositionPlotViewData } from './PositionPlotViewData'
+
+type Props = {
+    data: PositionPlotViewData
+    width: number
+    height: number
+}
+
+type PanelProps = {
+    pixelZero: number
+    dimensions: {
+        dimensionIndex: number
+        dimensionLabel: string
+        pixelTimes: number[]
+        pixelValues: number[]
+    }[]
+}
+
+const margins = {
+    left: 30,
+    right: 20,
+    top: 20,
+    bottom: 50
+}
+
+const panelSpacing = 4
+
+const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
+    const {startTimeSec, endTimeSec} = useMemo(() => {
+        const startTimeSec = min(data.timestamps)
+        const endTimeSec = max(data.timestamps)
+        return {startTimeSec, endTimeSec}
+    }, [data.timestamps])
+
+    // Compute the per-panel pixel drawing area dimensions.
+    const panelCount = 1
+    const { panelWidth, panelHeight } = useMemo(() => computePanelDimensions(width, height, panelCount, panelSpacing, margins), [width, height, panelCount])
+    const pixelsPerSecond = useMemo(() => computePixelsPerSecond(panelWidth, startTimeSec, endTimeSec), [endTimeSec, startTimeSec, panelWidth])
+
+    // We need to have the panelHeight before we can use it in the paint function.
+    // By using a callback, we avoid having to complicate the props passed to the painting function; it doesn't make a big difference
+    // but simplifies the prop list a bit.
+    const paintPanel = useCallback((context: CanvasRenderingContext2D, props: PanelProps) => {
+        context.strokeStyle = 'black'
+        context.setLineDash([5, 15]);
+        context.beginPath()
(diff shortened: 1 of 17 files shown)


magland:
Merge #5 first

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -0,0 +1,147 @@
...
+        for (let dim of props.dimensions) {
+            context.strokeStyle = colorForUnitId(dim.dimensionIndex)
+            context.beginPath()
+            let first = true
+            for (let i=0; i<dim.pixelTimes.length; i++) {
+                const x = dim.pixelTimes[i]
+                const y = dim.pixelValues[i]
+                if (first) context.moveTo(x, y)
+                else context.lineTo(x, y)
+                first = false
```
jsoules, line 65: Might be easier to just set the initial position outside the loop as `context.moveTo(dim.pixelTimes[0], dim.pixelValues[0])`.

src/views/PositionPlot/PositionPlotViewData.ts
```diff
@@ -0,0 +1,18 @@
+import { validateObject } from "figurl"
+import { isArrayOf, isEqualTo, isString } from "figurl/viewInterface/validateObject"
+
+export type PositionPlotViewData = {
+    type: 'PositionPlot'
+    timestamps: number[]
+    positions: number[][]
```
jsoules, line 7: My instinct is that some of the filtering code in PositionPlotView will be a bit more transparent if we make these into objects and index the objects, rather than having to keep the indices of several different lists synced up. (Not sure whether there'd be a noticeable performance impact to this.)

For the individual positions, is it safe to assume `positions[timestamp] => some_position[]` where  `some_position[0]` is the x-coordinate and `some_position[1]` is the y-coordinate? (And we probably don't anticipate a z-coordinate, but we could add it if we needed to?) It might be nice to map explicitly to named coordinate variables, but we would just have to map back to point vectors when we do the coordinate transform to get pixels, so I'm less fussed about that.

src/views/PositionPlot/PositionPlotViewData.ts
```diff
@@ -0,0 +1,18 @@
...
+    dimensionLabels: string[]
+}
+
+export const isPositionPlotViewData = (x: any): x is PositionPlotViewData => {
+    return validateObject(x, {
+        type: isEqualTo('PositionPlot'),
+        timestamps: () => (true),
+        positions: () => (true),
```
jsoules, line 15: Assuming one mark of invalid data would be if these two lists have different lengths.

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -0,0 +1,147 @@
...
+    }, [data.timestamps, startTimeSec, endTimeSec, data.dimensionLabels, data.positions])
+
+    const valueRange = useMemo(() => {
+        const yMin = Math.min(0, min(series.map(S => (min(S.values)))))
+        const yMax = Math.max(0, max(series.map(S => (max(S.values)))))
+        return {yMin, yMax}
+    }, [series])
+
+    const panels: TimeScrollViewPanel<PanelProps>[] = useMemo(() => {
+        const yMap = ((y: number) => (
```
jsoules, line 96: Wondering if this means now is the time to get around to implementing that two-dimensional transformation matrix function. (hmm)

jsoules: Actually--now I look at this in the gui, I'm not actually sure what's being plotted here. Time is on the x-axis, and the y-axis is the y-value of the animal's position? (Or some function of the x, y position?)

Also, looking at the output, I think we might want to think about ways to smooth or interpolate this data some.

jsoules (commented):
A couple suggested improvements, take or leave.
Happy for this to be merged, but I suspect we'll do some more iterating on the plot.

src/views/PositionPlot/PositionPlotView.tsx
magland: In some ways that's better (avoids the if), but it requires keeping 2 lines of code in sync (what if the definition of x,y changes). Also we get an extra segment in the path - I can't think of how this would harm us, but still. Leaving it as is for now.

src/views/PositionPlot/PositionPlotView.tsx
magland: The y value is the value of which ever coordinate we're plotting. And it's possible to plot more than one coordinate on the same graph. For example, here's the x- and y- coordinates plotted on the same graph:

image

jsoules: Ah, okay. As a non-expert, I don't have much intuition for what this graph is telling me, but if it makes sense to people working in the field, that's what matters.

Is the oscillation pattern a product of the data--like, the animal is actually rocking back and forth in some way? Or is it an artifact of sampling rates or something?

magland: The animal is running back and forth along a track. The timescale is such that each oscillation could be a minute or more.

jsoules: Oh, okay. With that scale in mind it makes a lot more sense.

PR #7, opened 2021-11-04
Curation auth

magland:
Implement authorization for sorting curation
Builds on PositionPlot PR

### DIFF
--- a/src/views/MountainLayout/MountainLayoutView.tsx
+++ b/src/views/MountainLayout/MountainLayoutView.tsx
@@ -1,8 +1,9 @@
 import SortingCurationAction from 'contexts/SortingCurationAction';
 import SortingCurationContext, { sortingCurationReducer } from 'contexts/SortingCurationContext';
-import { initiateTask, useSubfeedReducer } from 'figurl';
+import { initiateTask, useSignedIn, useSubfeedReducer } from 'figurl';
+import runTaskAsync from 'figurl/runTaskAsync';
 import MountainWorkspace from 'MountainWorkspace/MountainWorkspace';
-import React, { FunctionComponent, useCallback, useMemo } from 'react';
+import React, { FunctionComponent, useCallback, useEffect, useMemo, useState } from 'react';
 import { MountainLayoutViewData } from './MountainLayoutViewData';
 import ViewWrapper from './ViewWrapper';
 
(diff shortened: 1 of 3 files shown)


magland:
Merge #6 first

src/views/MountainLayout/MountainLayoutView.tsx
```diff
@@ -33,23 +34,49 @@ const MountainLayoutView: FunctionComponent<Props> = ({data, width, height}) =>
...
+            'spikesortingview.check_sorting_curation_authorized.1',
+            {
+                sorting_curation_uri: data.sortingCurationUri,
+                user_id: userId,
+                google_id_token: googleIdToken
+            },
+            'query',
+            {queryUseCache: false}
+        ).then((result) => {
+            // yeah, there's a race condition here
```
jsoules, line 73: On the assumption that the hard-false setting at l. 56 might complete after the (query-based) setting on line 74?
Could restructure as:
```
if (data.sortingCurationUri && userId && googleIdToken) {
  // do the check, return what it returns
} else {
  if (canCurate) setCanCurate(false)
}
```
maybe? Then we avoid triggering a state update (& consequent rerender) unless the state isn't already what we want it to be?

Or is the race condition something else?

jsoules (commented):
following up on the note of a race condition

src/views/MountainLayout/MountainLayoutView.tsx
magland: The problem is that if the userId changes while the check is happening, we could get the wrong value for canCurate. Since the check is async, there is no guarantee that the last check that completes corresponds to the current value of userId

jsoules: Ah! Okay. Could we do something like have `canCurate` return the userId that it checked, and then make a call to a function that has access to the current value or can otherwise bail if the response userId doesn't match the logged-in one? Like... supposing the userId has changed since the login, will the `.then` function have access to the *current* value, or is it closed over the value that was present when the async function originated?

(Or bigger-picture, I suppose this isn't strictly a security feature--just a convenience one so we don't show the curation interface if the logged-in user can't curate. So if it breaks, it just means we told the user they could do curations, and then when they try to persist them, it won't actually work. So I think we're also fine to just not worry about it.)

magland: It's not super easy to solve, but definitely possible. You can't just check the current state. And I agree with you that since it's  not a security issue, I vote that we don't worry about it. I don't expect it will be a problem because it's hard to imagine an actual case where it's likely to fail.

PR #10, opened 2021-11-09
Waveform standard deviations

magland:
Builds on #9 . Merge #9 first.

image


jsoules:
What's going on with Unit 4 in the screenshot?

magland:
Just a high amplitude waveform

jsoules:
> Just a high amplitude waveform

Okay. I see the waveform for electrode 2 has moved up to around electrode 0, which seemed rather... intense. Just making sure it's just the data and not that we've got something wrong in the coordinate mappings.

PR #12, opened 2021-11-15
Recording selection context jfm

magland:
* change focus time to click, even if not focused
* TSV zoom: adjust behavior to work well with trackpad
* Implement mouse panning

### DIFF
--- a/src/contexts/RecordingSelectionContext.ts
+++ b/src/contexts/RecordingSelectionContext.ts
@@ -212,44 +212,12 @@ const initializeRecordingSelection = (state: RecordingSelection, action: Initial
     return newState
 }
 
-const panTime = (state: RecordingSelection, action: PanRecordingSelectionAction): RecordingSelection => {
+const panTimeHelper = (state: RecordingSelection, panDisplacementSeconds: number) => {
     if (state.visibleTimeStartSeconds === false || state.visibleTimeEndSeconds === false || state.recordingStartTimeSeconds === false || state.recordingEndTimeSeconds === false) {
         console.warn(`WARNING: Attempt to call panTime() with uninitialized state ${state}.`)
         return state
     }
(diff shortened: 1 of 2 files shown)


magland:
Live view: https://www.figurl.org/f?v=gs://figurl/spikesortingview-1&d=ae21d3690de8b27d269eda21c4d64d62e1eaa910&channel=flatiron1&label=Test%20MV%20layout

src/contexts/RecordingSelectionContext.ts
```diff
@@ -224,6 +243,36 @@ const panTime = (state: RecordingSelection, action: PanRecordingSelectionAction)
...
+        return state
+    }
+    const keepFocus = state.focusTimeSeconds && state.focusTimeSeconds > newStart && state.focusTimeSeconds < newEnd
+    const focus = keepFocus ? state.focusTimeSeconds : undefined
+
+    // Avoid creating new object if we didn't actually change anything
+    if (newStart === state.visibleTimeStartSeconds && newEnd === state.visibleTimeEndSeconds) return state
+
+    // console.log(`Returning new state: ${newStart} - ${newEnd} (was ${state.visibleTimeStartSeconds} - ${state.visibleTimeEndSeconds})`)
+    return {...state, visibleTimeStartSeconds: newStart, visibleTimeEndSeconds: newEnd, focusTimeSeconds: focus }
```
jsoules, line 273: Since this logic is repeated for both state changes (pan by pct and pan by defined amount) it might be a good idea to pop it out into a function that both verbs call.

jsoules (commented):
This looks good, just one point where there's some repeated logic that we should probably make a function.

src/contexts/RecordingSelectionContext.ts
magland: I created a helper function for this as you suggested

PR #14, opened 2021-11-16
implement time axis ticks

magland:
Implemented time axis ticks (not really ticks, but vertical lines with time labels)

The time labels are relative... for example
Time `1day 3hr 45min 17s 113ms` would just be labeled as 113 ms (if this were a msec label)
Or if it were a minute label:  `1day 3hr 45min` would be labeled as 45 min

I think this works pretty well, but the user would also like to know the absolute time... so I think that could be achieved by showing the full time marker for the focus time (that would appear elsewhere in some text box)

There are major and minor "ticks", and there are some heuristics for determining which tick level to display -- and this adjusts based on the width of the view.

A screenshot:

image

### DIFF
--- a/src/views/RasterPlot/TimeScrollView/TSVAxesLayer.tsx
+++ b/src/views/RasterPlot/TimeScrollView/TSVAxesLayer.tsx
@@ -1,11 +1,12 @@
 import BaseCanvas from 'FigurlCanvas/BaseCanvas';
 import React, { useMemo } from 'react';
 import { paintAxes } from './paint';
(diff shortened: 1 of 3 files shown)


jsoules:
I think this looks really good, but I'm going to suggest handling the tick computation as a hook provided by `TimeScrollView` rather than having it defined in the paint functions. That would also let us get around the need to do scaling operations in the paint function (since we can set the pixel-x-coordinate of each tick within that hook). I'm also going to suggest making some of the field names a little more descriptive.

Will push a PR to this branch in a sec with those changes. I apologize if they seem excessive!

PR #15, opened 2021-11-16
Use hook for time tick computation

jsoules:
This moves the time tick computation into a memoized hook in `TimeScrollView.tsx` and renames some of the variables. It also uses vectorized pixel placement computation at the time the ticks are computed, to avoid doing it in the paint function (even though that's probably not a big driver of anything under present expectations).

To account for the margins, I added an optional (defaults to 0) extra pixel offset for the 1d time to pixel matrix computations; this fixed a bug I introduced while playing with this.


magland:
Looks good to me, merging.

PR #18, opened 2021-12-10
Position PDF plot

magland:


magland:
Implemented the position PDF plot and the epochs view as in this example:
https://www.figurl.org/f?v=gs://figurl/spikesortingview-1&d=2ae09cc57e2273ab2efa5285c8619cda338a82ae&channel=flatiron1&label=Position%20PDF%20and%20position

magland:
Also in the timescrollview, separated out the cursor layer from the axis layer, and put the cursor layer on top so that it is visible when the main plot obscures the view (e.g., in the position PDF plot)

src/views/PositionPdfPlot/useFetchCache.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (state.activeFetches[h]) return
+        dispatch({type: 'startFetch', queryHash: h})
+        fetchFunction(query).then((data: any) => {
+            if (data !== undefined) {
+                dispatch({type: 'setData', queryHash: h, data})
+            }
+        }).catch((err) => {
+            console.warn(err)
+            console.warn('Problem fetching data', query)
+            // note: we intentionally do not unset the active fetch here
```
jsoules, line 109: That's to keep from repeatedly firing the fetch?

src/views/PositionPdfPlot/useFetchCache.ts
```diff
@@ -0,0 +1,125 @@
...
+            throw Error('Unexpected action in fetchCacheReducer')
+        }
+    }
+}
+
+const queryHash = <QueryType>(query: QueryType) => {
+    return objectHash(query)
+}
+
+const useFetchCache = <QueryType>(fetchFunction: (query: QueryType) => Promise<any>): FetchCache<QueryType> => {
```
jsoules, line 72: The comments really helped make this clearer. Thanks!

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -0,0 +1,229 @@
...
+        // if (downsampleFactor > 9) return undefined
+        const ret = allocate2d(i2 - i1, numPositions, undefined)
+        for (let s = s1; s < s2; s++) {
+            const S = fetchSegmentCache.get({type: 'fetchSegment', segmentNumber: s, segmentSize, downsampleFactor})
+            if (S) {
+                for (let i = 0; i < S.length; i++) {
+                    const j = s * segmentSize - i1 + i
+                    if ((0 <= j) && (j < i2 - i1)) {
+                        for (let p = 0; p < numPositions; p++) {
+                            ret[j][p] = S[i][p]
```
jsoules, line 53: This is summing the vectors but there's no divide step for the average--so you'd wind up with just a sum--is this intentional?

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -0,0 +1,229 @@
...
+    const fetchSegmentCache = useFetchCache<FetchSegmentQuery>(fetchSegment)
+    const get = useCallback((i1: number, i2: number, downsampleFactor: number) => {
+        const s1 = Math.floor(i1 / segmentSize)
+        const s2 = Math.ceil(i2 / segmentSize)
+        if (s2 - s1 + 1 >= 8) return undefined
+        // if (downsampleFactor > 9) return undefined
+        const ret = allocate2d(i2 - i1, numPositions, undefined)
+        for (let s = s1; s < s2; s++) {
+            const S = fetchSegmentCache.get({type: 'fetchSegment', segmentNumber: s, segmentSize, downsampleFactor})
+            if (S) {
```
jsoules, line 48: I think my suggested alterations to the downsampling loop in `PositionPdfPlotView` might be applicable here too.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }, [visibleValues])
+
+    const imageData = useMemo(() => {
+        if (!visibleValues) return undefined
+        if (minValue === undefined) return undefined
+        if (maxValue === undefined) return undefined
+        const N1 = visibleValues.length
+        const N2 = visibleValues[0].length
+        const imageData = new ImageData(N1, N2)
```
jsoules, line 118: What's an `ImageData`? It doesn't seem to be defined or imported anywhere.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -0,0 +1,229 @@
...
+                }
+                else {
+                    imageData.data[i + 0] = 0
+                    imageData.data[i + 1] = 0
+                    imageData.data[i + 2] = 0
+                    imageData.data[i + 3] = 0
+                }
+                i += 4
+            }
+        }
```
jsoules, line 139: This is a bit terse. Looks like you're making a heatmap by converting the vector components of the underlying pdf data object (as downsampled) into a proportion of their range, and then writing that into the RGB channels, something like that?

Using `i2` for the outer loop iterator and `i1` for the inner loop iterator is not the most conventional choice :)
The outer loop is iterating over which component of the pdf you're encoding, and the inner one is iterating over the time slot, do I have that right? I'm assuming this has to do with the layout of the `ImageData` in memory and how it corresponds to the rows/columns of pixels in the underlying image object?

jsoules (commented):
This looks good overall, but I think a few parts could benefit from some more explanation. I'm posting a PR with suggestions for clarifying one of the downsampling loops--didn't want to get too deep into the other two parts that I didn't completely follow (those being the downsampling and the image plotting in `PositionPdfPlotWidget`).

src/views/Epochs/EpochsView.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    const paintPanel = useCallback((context: CanvasRenderingContext2D, props: PanelProps) => {
+        context.clearRect(0, 0, panelWidth, panelHeight)
+        for (let E of pixelEpochs) {
+            context.fillStyle = 'lightgray'
+            context.strokeStyle = 'gray'
+            context.fillRect(E.startPixel, 0, E.endPixel - E.startPixel, panelHeight)
+            context.strokeRect(E.startPixel, 0, E.endPixel - E.startPixel, panelHeight)
+            context.fillStyle = 'black'
+            context.fillText(E.epoch.label, Math.max(E.startPixel, 0) + 10, 20)
```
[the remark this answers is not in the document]
magland: Okay good to know. Yeah, this case is not render-heavy at all.

src/views/PositionPdfPlot/PositionPdfPlotView.tsx
```diff
@@ -0,0 +1,54 @@
...
+import PositionPdfPlotWidget, { allocate2d, FetchSegmentQuery } from './PositionPdfPlotWidget'
+
+type Props = {
+    data: PositionPdfPlotViewData
+    width: number
+    height: number
+}
+
+const segmentSize = 10000
+const multiscaleFactor = 3
```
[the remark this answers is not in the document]
magland: I think it's okay to hard-code these for now. This is the case of a static figure. Note that for a live figure, the multiscale factor and segment size come in as part of the dataset.

src/views/PositionPdfPlot/PositionPdfPlotView.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const PositionPdfPlotView: FunctionComponent<Props> = ({data, width, height}) => {
+    const numPositions = data.pdf[0].length
+    const fetchSegment = useMemo(() => (async (query: FetchSegmentQuery) => {
+        const ret = allocate2d(query.segmentSize, numPositions, 0) as number[][]
+        for (let i = 0; i < query.segmentSize; i++) {
```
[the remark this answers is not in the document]
magland: Okay thanks

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -0,0 +1,229 @@
...
+}
+
+const panelSpacing = 4
+
+const usePositionPdfDataModel = (fetchSegment: (q: FetchSegmentQuery) => Promise<number[][]>, numTimepoints: number, numPositions: number, segmentSize: number, multiscaleFactor: number) => {
+    const fetchSegmentCache = useFetchCache<FetchSegmentQuery>(fetchSegment)
+    const get = useCallback((i1: number, i2: number, downsampleFactor: number) => {
+        const s1 = Math.floor(i1 / segmentSize)
+        const s2 = Math.ceil(i2 / segmentSize)
+        if (s2 - s1 + 1 >= 8) return undefined
```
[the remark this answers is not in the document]
magland: This is old from something I was trying... I will remove these numbers and clean up this part of the code.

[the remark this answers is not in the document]
magland: This has now been removed.

src/views/PositionPdfPlot/useFetchCache.ts
```diff
@@ -0,0 +1,125 @@
...
+            dispatch({type: 'clear'})
+        }
+    }, [fetchFunction])
+    const get = useMemo(() => ((query: QueryType) => {
+        const h = queryHash(query)
+        const v = state.data[h]
+        if ((v === undefined) && (!state.activeFetches[h])) {
+            if (!queriesToFetch.current[h]) {
+                queriesToFetch.current[h] = query
+                setCount((c) => (c + 1)) // make sure we trigger a state change so we go to the useEffect below
```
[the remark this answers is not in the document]
magland: I think so, it's been quite a while since I wrote this. It was tricky to get right... at some point we may want to rewrite this more clearly.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
magland: There is no summing here. It's just copying the data.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
magland: It's native data object that can be drawn to an HTML5 canvas directly -- sent directly to the GPU.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
magland: Yes. The reason the loops are nested the way they are is because we need to write imageData in a specific order.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
jsoules: Ah, yes, so it is.

PR #20, opened 2022-01-19
Addresses SortingView issues 168, 169.

jsoules:
This PR includes several things, some of which are necessary and some of which might be more controversial.

It's intended to address SortingView issues [168](https://github.com/magland/sortingview/issues/168) (support markers vs connecting points for linearized position plot) and [169](https://github.com/magland/sortingview/issues/169) (lines cutting out when the observation value is the same for many observations), which request changes to the Linear Position plots (where the functional component is the [PositionPlotView.tsx](https://github.com/magland/spikesortingview/blob/2eeb6e6ff8102077fda4edb74841e2296015c689/src/views/PositionPlot/PositionPlotView.tsx)).

Fixed line mode:
image

Scatterplot:
image

(Note the scatterplot reveals discontinuities in the data.)

Key changes:

- Scatterplot vs Linear mode is governed by the data type, which is used by `View.tsx` to choose which component will be used to display the data. I added `PositionPlotScatter` as a permitted data type that will map to this component. It would also have been possible to implement this by adding an additional value to the `data` field passed in--we might yet choose to implement it that way (it was how I'd originally done it) but I was hoping to limit the number of places where we mix actual data and metadata that is actually display logic.
- In `PositionPlotView`, I've added scatterplot support to the `paintPanel` function.
- Additionally, tried setting line width for line mode to 1.1, which eliminated the 'vanishing line' issue (169).
- Not a key change, but a probably useful one--I also added a `getYAxisPixelZero` function to `TimeScrollView.tsx` that, given a transformation matrix, identifies the baseline of a plot. We seem to do this a lot, so might as well functionalize it.

Other changes:
These are code rewrites that were not directly required by the changes that directly solve the issues--we might choose to back them out. But I think they improve clarity and efficiency & make the code more idiomatic.

- In the scatterplot painting function, I iterate using a `forEach` over `props.dimensions` and then another over the `pixelTimes`/`pixelValues`; I also rewrote the line function to use this style instead of explicit `for` loops.
- In building the data series, I reworked the time filtering to filter times outside of the per-dimension loop & take the indices represented by the filtered times.
- In converting data to pixelspace for drawing on the panels, I switched to using the vectorized/matrix-based data point conversion using `use2dPanelDataToPixelMatrix` from `TimeScrollView` rather than the explicit function version. This ought to be more efficient (because of the vectorization), but it could be made even more so in two ways. First, I'm doing one matrix multiplication per dimension, rather than packing all the dimensions, multiplying, and unpacking. (This is probably not a big loss if we only have two or three dimensions.) Second, given that each of the dimensions shares the same time values, we could avoid converting the times repeatedly. However, this would require an additional 1d conversion matrix function in `TimeScrollView` in addition to the time-conversion function and time-and-data conversion function, which I didn't bother with.

To actually see the scatter plot in practice, we'll probably need to redo the FigURL so that it takes advantage of scatterplot mode where appropriate. But for review purposes, you can just uncomment the line (l. 149 in `PositionPlotView`) that manually makes everything a scatterplot.

### DIFF
--- a/src/views/PositionPlot/PositionPlotView.tsx
+++ b/src/views/PositionPlot/PositionPlotView.tsx
@@ -2,7 +2,7 @@ import { useRecordingSelectionTimeInitialization, useTimeRange } from 'contexts/
 import { matrix, multiply } from 'mathjs'
 import React, { FunctionComponent, useCallback, useMemo } from 'react'
 import colorForUnitId from 'views/common/colorForUnitId'
-import TimeScrollView, { TimeScrollViewPanel, use1dTimeToPixelMatrix, usePanelDimensions, usePixelsPerSecond } from '../RasterPlot/TimeScrollView/TimeScrollView'
+import TimeScrollView, { getYAxisPixelZero, TimeScrollViewPanel, use2dPanelDataToPixelMatrix, usePanelDimensions, usePixelsPerSecond } from '../RasterPlot/TimeScrollView/TimeScrollView'
 import { PositionPlotViewData } from './PositionPlotViewData'
 
 type Props = {
@@ -19,6 +19,7 @@ type PanelProps = {
         pixelTimes: number[]
         pixelValues: number[]
     }[]
+    plotType: 'line' | 'scatter'
 }
 
 const margins = {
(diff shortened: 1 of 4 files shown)


src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
```diff
@@ -145,6 +145,12 @@ export const use2dPanelDataToPixelMatrix = (pixelsPerSecond: number, startTimeSe
     }, [pixelsPerSecond, startTimeSec, dataMin, dataMax, userScaleFactor, panelHeight, invertY])
 }
 
+export const getYAxisPixelZero = (TransformMatrix2d: Matrix) => {
+    const augmentedZero = matrix([[0], [0], [1]])
+    const value = (multiply(TransformMatrix2d, augmentedZero).valueOf() as number[][])[1][0]
+    return value
+}
+
```
jsoules, line 153: Utility function for finding the pixel value of the zero point along a y-axis. (Probably most useful if the axis is inverted.)

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -90,30 +114,43 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
...
-        const yMap = ((y: number) => (
-            (1 - (y - valueRange.yMin) / (valueRange.yMax - valueRange.yMin)) * panelHeight
-        ))
+        const pixelZero = getYAxisPixelZero(pixelTransform)
+        // this could also be done as one matrix multiplication by concatenating the dimensions;
+        // and we could even separate out the time series values (which are repeated). But probably not worth it.
+        // TODO: ought to profile these two versions
+        const pixelData = series.map(s => {
+            const augmentedPoints = matrix([ s.times, s.values, new Array(s.times.length).fill(1) ])
+            const pixelPoints = multiply(pixelTransform, augmentedPoints).valueOf() as number[][]
```
jsoules, line 134: This uses the 2d data-to-pixel transformation matrix provided by `TimeSeriesView.use2dPanelDataToPixelMatrix()` (called above) to vectorize the conversion from dataspace to pixelspace. Technically we could reduce the amount of computation by a) reshaping the data so the dimensions are all done at once, and b) operating on the (repeated) time elements once, and then converting only the y-axis data, but a) seems like more trouble than it's worth for small numbers of dimensions and b) would require another conversion-matrix hook plus chopping up the data more, so I've left it alone for now. (I give myself permission by noting that the existing version is also converting the times into pixels once per dimension.)

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -52,35 +53,58 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
...
+    // const timeToPixelMatrix = use1dTimeToPixelMatrix(pixelsPerSecond, visibleTimeStartSeconds)
 
     const series = useMemo(() => {
         const series: {dimensionIndex: number, times: number[], values: number[]}[] = []
-        for (let dimensionIndex=0; dimensionIndex<data.dimensionLabels.length; dimensionIndex++) {
-            const filteredTimes = data.timestamps.filter(t => (visibleTimeStartSeconds <= t) && (t <= visibleTimeEndSeconds))
-            const filteredValues = data.timestamps.map((t, ii) => (data.positions[ii][dimensionIndex])).filter((a, ii) => (visibleTimeStartSeconds <= data.timestamps[ii]) && (data.timestamps[ii] <= visibleTimeEndSeconds))
+        const filteredTimeIndices = data.timestamps.flatMap((t, ii) => (visibleTimeStartSeconds <= t) && (t <= visibleTimeEndSeconds) ? ii : [])
+        const filteredTimes = filteredTimeIndices.map(i => data.timestamps[i])
+        const filteredValues = filteredTimeIndices.map(index => data.positions[index])
```
jsoules, line 89: Here I'm getting the filtered indices from the time series once, and then populating times and values from those indices. This avoids the need to filter on each loop & makes the logic a bit more straightforward vs. needing to filter one array based on looking up values in another.

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -52,35 +53,58 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
...
-            context.stroke()
+        if (props.plotType === 'line' || !props.plotType) {
+            props.dimensions.forEach(dim => {
+                context.strokeStyle = colorForUnitId(dim.dimensionIndex)
+                context.lineWidth = 1.1 // hack--but fixes the 'disappearing lines' issue
+                context.beginPath()
+                dim.pixelTimes.forEach((x, ii) => {
+                    const y = dim.pixelValues[ii]
+                    ii === 0 ? context.moveTo(x, y) : context.lineTo(x, y)
+                })
```
jsoules, line 64: Using `forEach` with an index argument, & using "index is 0" to test for the first pass rather than the helper variable.

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -52,35 +53,58 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
...
+                context.stroke()
+            })
+        } else if (props.plotType === 'scatter') {
+            const circleWidthPx = 0.1 // could go even lighter
```
jsoules, line 68: I played around with this a little bit. It might become a parameter or something?

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -52,35 +53,58 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
...
-                const y = dim.pixelValues[i]
-                if (first) context.moveTo(x, y)
-                else context.lineTo(x, y)
-                first = false
-            }
-            context.stroke()
+        if (props.plotType === 'line' || !props.plotType) {
+            props.dimensions.forEach(dim => {
+                context.strokeStyle = colorForUnitId(dim.dimensionIndex)
+                context.lineWidth = 1.1 // hack--but fixes the 'disappearing lines' issue
```
jsoules, line 59: Not a huge fan of "magic numbers" like this, but it fixes the vanishing-line issue, at least for the example we have been looking at.

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -52,35 +53,58 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
...
+                dim.pixelTimes.forEach((x, ii) => {
+                    const y = dim.pixelValues[ii]
+                    ii === 0 ? context.moveTo(x, y) : context.lineTo(x, y)
+                })
+                context.stroke()
+            })
+        } else if (props.plotType === 'scatter') {
+            const circleWidthPx = 0.1 // could go even lighter
+            props.dimensions.forEach(d => {
+                context.fillStyle = colorForUnitId(d.dimensionIndex)
```
jsoules, line 70: Side thought--we could probably do away with the need for the `dimensionIndex` member altogether if we just take the index of `props.dimensions`. Not sure if there's any other significance to having the dimension index explicitly recorded in the data structure.

jsoules (commented):
Turns out I'm mostly commenting on the changes that *aren't* required...!

magland:
This looks good. I am merging and then will publish to the figurl bucket.

PR #22, opened 2022-01-20
PositionPlotScatter: use fillRect rather than filling a circle - faster rendering

### LINKED ISSUE #21: PositionPlotScatter: use fillRect rather than filling a circle - faster rendering

magland:
Fixes #21

### DIFF
--- a/src/views/PositionPlot/PositionPlotView.tsx
+++ b/src/views/PositionPlot/PositionPlotView.tsx
@@ -65,13 +65,19 @@ const PositionPlotView: FunctionComponent<Props> = ({data, width, height}) => {
                 context.stroke()
             })
         } else if (props.plotType === 'scatter') {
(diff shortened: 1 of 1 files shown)


magland:
@jsoules, we can still keep tracking issues on sortingview repo. But this one was internal to the spikesortingview code, so I filed it here. No hard and fast rule, I suppose.

jsoules (approved):
Looks great. I would go ahead and delete the commented-out code, we can find it easily enough if we need.

Out of curiosity, does this still work if we do decimal values for the width/height?

magland:
> Out of curiosity, does this still work if we do decimal values for the width/height?

I tried it and yes it 'works'. However then some of the markers are not visible, depending on their subpixel location.

I went ahead and deleted the commented code. I also shifted back by 0.5 pixels so that the 1-pixel rect is centered at the correct location.

PR #23, opened 2022-01-25
implement multi timeseries view

magland:


magland:
When timeseries figures are panels in a MultiTimeseriesView, the left toolbar is hidden, the time axis is hidden (except for the bottom one), and the label is displayed as rotated vertical text to the left of the timeseries.

image

src/views/MultiTimeseries/MultiTimeseriesView.tsx
```diff
@@ -0,0 +1,40 @@
...
+    width: number
+    height: number
+}
+
+const MultiTimeseriesView: FunctionComponent<Props> = ({data, width, height}) => {
+    const divStyle: React.CSSProperties = useMemo(() => ({
+        width,
+        height,
+        overflowX: 'hidden',
+        overflowY: 'auto'
```
jsoules, line 16: I think it might be easier to trace things out later if these two static properties were set by a CSS class (something like `.MultiTimeseriesViewWrapperDiv` or whatever) rather than coming from code.

src/views/MultiTimeseries/ViewWrapper.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    const labelDivStyle: React.CSSProperties = useMemo(() => ({
+        position: 'absolute',
+        writingMode: 'vertical-lr',
+        transform: 'rotate(-180deg)',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+        left: 0,
+        top: 0,
```
jsoules, line 76: Again, for the non-computed CSS properties I think it'll be easier to follow in the future if we make these part of a semantically-named class and apply that where appropriate.

src/views/MultiTimeseries/MultiTimeseriesView.tsx
```diff
@@ -0,0 +1,40 @@
...
+    }), [width, height])
+
+    const H = height / data.panels.length
```
jsoules, line 19: Does this need to take into account the relative-height?

src/views/Epochs/EpochsView.tsx
```diff
@@ -20,21 +22,17 @@ type PanelProps = {
...
 const EpochsView: FunctionComponent<Props> = ({data, width, height}) => {
     const {visibleTimeStartSeconds, visibleTimeEndSeconds } = useTimeRange()
 
+    const margins = useTimeseriesMargins(undefined)
+
     // Compute the per-panel pixel drawing area dimensions.
     const panelCount = 1
-    const { panelWidth, panelHeight } = usePanelDimensions(width, height, panelCount, panelSpacing, margins)
+    const toolbarWidth = DefaultToolbarWidth
+    const { panelWidth, panelHeight } = usePanelDimensions(width - toolbarWidth, height, panelCount, panelSpacing, margins)
```
jsoules, line 35: Upon reflection (looking at this & at the `PositionPdfPlotWidget` below), I think it might be easier to add a `toolbarWidth` parameter to the `usePanelDimensions` hook (perhaps an optional one defaulting to `DefaultToolbarWidth`?) so that you don't have to remember to subtract it out when using the hook.

jsoules: Or a `hideToolbar` boolean or something?

src/views/RasterPlot/TimeScrollView/paint.ts
```diff
@@ -18,24 +18,29 @@ const highlightedRowFillStyle = '#c5e1ff' // TODO: This should be standardized a
...
+        context.strokeStyle = 'black'
+        drawLine(context, margins.left, height - margins.bottom, width - margins.right, height - margins.bottom)
+    }
 
     // time ticks
     for (let tt of timeTicks) {
         // const frac = (tt.value - timeRange[0]) / (timeRange[1] - timeRange[0])
         // const x = margins.left + frac * (width - margins.left - margins.right)
         context.strokeStyle = tt.major ? 'gray' : 'lightgray'
         drawLine(context, tt.pixelXposition, height - margins.bottom, tt.pixelXposition, margins.top)
```
jsoules, line 35: Oh, right, this is the code that draws the vertical separator lines that are visible across the time series.

src/views/RasterPlot/TimeScrollView/paint.ts
```diff
@@ -18,24 +18,29 @@ const highlightedRowFillStyle = '#c5e1ff' // TODO: This should be standardized a
...
-        context.textAlign = 'center'
-        context.textBaseline = 'top'
-        const y1 = height - margins.bottom + 5
-        context.fillStyle = tt.major ? 'black' : 'gray'
-        context.fillText(tt.label, tt.pixelXposition, y1)
+        if (!hideTimeAxis) {
+            drawLine(context, tt.pixelXposition, height - margins.bottom, tt.pixelXposition, height - margins.bottom + 5)
```
jsoules, line 37: And *this* is the line that draws the actual tick marks, which are only drawn in if the axis is displayed. Got it.
Might be worthwhile to throw a comment in to identify which lines are being drawn by these two steps.

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
```diff
@@ -444,55 +448,66 @@ const TimeScrollView = <T extends {[key: string]: any}> (props: TimeScrollViewPr
...
-            initialPosition={DefaultToolbarWidth}
+            initialPosition={toolbarWidth}
             adjustable={false}
         >
             <ViewToolbar
-                width={DefaultToolbarWidth}
+                width={toolbarWidth}
                 height={height}
                 customActions={timeControlActions}
             />
```
jsoules, line 509: Does this mean the toolbar `div` still gets rendered into the DOM, even if the toolbar is hidden? Is that to make sure the keyboard/mouse controls still work, or something like that?

jsoules: Seems to me kind of like we'd want to just ditch the whole `Splitter` if there's no toolbar, but I haven't really thought about it at all.

src/views/PositionPlot/PositionPlotView.tsx
```diff
@@ -22,23 +25,45 @@ type PanelProps = {
...
+export const useTimeseriesMargins = (timeseriesLayoutOpts: TimeseriesLayoutOpts | undefined) => {
+    return useMemo(() => {
+        const {hideTimeAxis, hideToolbar} = timeseriesLayoutOpts || {}
+        if (hideToolbar) {
+            return (
+                {
+                    left: 30,
+                    right: 20,
+                    top: 20,
+                    bottom: hideTimeAxis ? 20 : 50
```
jsoules, line 39: Noting the difference in bottom margin based on `hideTimeAxis`--where toolbar + axis gets us 50 pixels but toolbar - axis gets us 40. Is this because the `hideToolbar` (or `hideTimeAxis`) parameter is also driving where the label goes, is that right?

jsoules (commented):
Here's some feedback based on first read-through. I haven't had the chance to pull it and look at the live examples yet, but I'll do that in the morning.

Overall this looks great! I just had a couple suggestions and a couple questions to make sure I'm following everything.

src/View.tsx
```diff
@@ -50,16 +60,16 @@ const View: FunctionComponent<Props> = ({data, width, height}) => {
         return <ElectrodeGeometryView data={data} width={width} height={height} />
     }
     else if (data.type === 'PositionPlot' || data.type === 'PositionPlotScatter') {
```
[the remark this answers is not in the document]
magland: Maybe I'm using the term "timeseries" incorrectly here. But TimeseriesLayoutOpts is supposed to apply to any plot that has a time axis. This for example includes spike raster plot, spike amplitude plot, position probability, etc. So it wouldn't be an appropriate place for options that are specific to one type of view (e.g., position plot)

src/views/Epochs/EpochsView.tsx
```diff
@@ -1,6 +1,8 @@
 import { useTimeRange } from 'contexts/RecordingSelectionContext'
 import { matrix, multiply } from 'mathjs'
 import React, { FunctionComponent, useCallback, useMemo } from 'react'
+import { DefaultToolbarWidth } from 'views/common/TimeWidgetToolbarEntries'
+import { useTimeseriesMargins } from 'views/PositionPlot/PositionPlotView'
```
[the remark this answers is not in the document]
magland: yes it certainly does. Let's do that in a separate PR.

src/views/MultiTimeseries/MultiTimeseriesView.tsx
magland: yeah that would be the cleaner way to do it.

src/views/MultiTimeseries/ViewWrapper.tsx
magland: yeah that would be the cleaner way to do it.

src/views/MultiTimeseries/MultiTimeseriesView.tsx
magland: yes it does. not implemented yet. Let's do that in a separate PR

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
magland: if you look at the code above the highlighted lines, you'll see that the splitter is indeed bypassed when toolbar is hidden.

src/views/PositionPlot/PositionPlotView.tsx
magland: Yeah the choices of numbers here is not super clear. If the toolbar is hidden, it means we are trying to save space, so I decrease the margins all around.

src/views/RasterPlot/TimeScrollView/paint.ts
magland: Okay, I'll put in a comment.

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
jsoules: I saw that the div ref is being used for the content, but it looked like the return still includes the splitter & toolbar divs--I wasn't quite sure how React would handle the situation where multiple divs use the same ref.

So I checked the live version and it looks like the splitter is still getting rendered with a toolbar child as well as the plot, it's just the toolbar width gets set to 0.

This doesn't seem to affect anything functionally, it just makes the page code more complicated--do we care?

magland: Oops you're right. I forgot to include the "return" keyword in the case of hiding the toolbar. Will fix.

PR #25, opened 2022-01-26
support timeseriesLayoutOpts in all timeseries views

### LINKED ISSUE #24: update raster plot to be a multitimeseries panel

magland:
update raster plot and other views to be
compatible with multitimeseries panel
Fixes #24


magland:
@jsoules this one is pretty straightforward. Just adapting other timeseries views to support timeseriesLayoutOpts

jsoules (commented):
All seems reasonable.

PR #26, opened 2022-01-26
Position-plot-discontinuous

magland:
Merge #25 first


jsoules (approved):
I agree, this is a nice way to control the rendering while keeping the `data` object describing the data itself. Looks good.

PR #30, opened 2022-02-04
Fixes #29. Implements relative heights in MultiTimeseriesView.

### LINKED ISSUE #29: Make MultiTimeseriesView honor relative-height requests
See src/views/MultiTimeseries/MultiTimeseriesView.tsx

We currently track the requested relative height of each panel, but we're dividing the space into N even slices for N panels regardless of the relative height parameters.

Probably need to set the total effective height as the sum of the relative heights for all panels, then adjust the allocated space for each panel according to its relative height.

The big trick here will be keeping track of how much relative height has been allocated so far--the paint method (https://github.com/magland/spikesortingview/blob/087735db9c8eaeaa38ae33839861bb51cb2b6e71/src/views/RasterPlot/TimeScrollView/paint.ts#L4-L14) currently assumes each panel has the same height and is using a vertical offset based on panel count. Instead, we'll need to make this method aware that panels can have different heights, so that it offsets by the sum of the relative-height-adjusted count of 'panel heights' already allocated.

i.e. for N panels N_1...N_n, where each panel has relative height h_i, total 'allocated height' .
https://github.com/magland/spikesortingview/blob/087735db9c8eaeaa38ae33839861bb51cb2b6e71/src/views/MultiTimeseries/MultiTimeseriesView.tsx#L19 should find the per-panel-height H_pp = height / H_t

and
https://github.com/magland/spikesortingview/blob/087735db9c8eaeaa38ae33839861bb51cb2b6e71/src/views/MultiTimeseries/MultiTimeseriesView.tsx#L24-L33
rendering each panel should use `height={panel.relativeHeight*H_pp}` so the relative height gets applied.

In https://github.com/magland/spikesortingview/blob/087735db9c8eaeaa38ae33839861bb51cb2b6e71/src/views/RasterPlot/TimeScrollView/paint.ts#L4-L14
we need to adjust the `panels` property so it tracks per-panel height, and then change the line
`context.translate(0, perPanelOffset)`
to `context.translate(0, panels[i].realizedPanelHeight)`
or something like that. (Don't recall if we optionally add a margin between panels here...)

To verify:
Example https://www.figurl.org/f?v=gs://figurl/spikesortingview-1-dev&d=601993ad3b8bc6975b8eb7c98c1c242d0276b453&channel=flatiron1&label=Jaq_03_12_visualization_data should show no difference from live version
NEW FIGURL EXAMPLE NEEDED, derived from this example, will set 'Position probability' panel to 2x height, and should show half the total drawing space allocated to the top two plots, and the other half of the drawing space allocated to the 'Position probability' plot.

jsoules:
Added relative heights for panels in multi-panel-view display.
Examples:
(1) https://www.figurl.org/f?v=gs://figurl/spikesortingview-1-dev&d=601993ad3b8bc6975b8eb7c98c1c242d0276b453&channel=flatiron1&label=Jaq_03_12_visualization_data
(2) https://www.figurl.org/f?v=gs://figurl/spikesortingview-1&d=964498c2fa87ebd882cf4aa8db7e36486fbc491d&channel=flatiron1&label=Jaq_03_12_visualization_data

Example (1) has been used previously to demonstrate the MultiTimeseriesView. No relative heights are set for any panels in this example. Comparing the code in the present PR with the live version of this FigURL suggests there are no regressions--the code as implemented still works like currently if there are no relative heights specified (since every panel defaults to a relative height of 1).

Example (2) shows relative heights being honored. This example gives a relative height of 4 to the 'Spikes' plot, 2 to the 'Linear Position' plot, none defined to the 'Speed' plot, and 3 to the 'Position Probability' plot. This results in the overall space being split into 10 equal slices, allocated as 4 to the first panel, 2 to the second panel, 1 to the third panel, and 3 to the fourth panel (as requested). Panel heights add up to slightly less than the overall space, because I implemented a `floor()` operation to avoid rounding-error push-up; this probably doesn't matter either way.

Example 2, before this change:
image

Example 2, after this change:
image

Note, example (2) from `sortingview-dev` repo's `/examples/multipanel_example_2.py`, which is part of PR #1 for https://github.com/scratchrealm/sortingview-dev/pull/1.

### DIFF
--- a/src/views/MultiTimeseries/MultiTimeseriesView.tsx
+++ b/src/views/MultiTimeseries/MultiTimeseriesView.tsx
@@ -16,7 +16,9 @@ const MultiTimeseriesView: FunctionComponent<Props> = ({data, width, height}) =>
         overflowY: 'auto'
     }), [width, height])
 
-    const H = height / data.panels.length
+    const total_height_allocated = data.panels.reduce((total, panel) => total + (panel?.relativeHeight ?? 1), 0)
+    const unit_height = Math.floor(height / total_height_allocated)
+    console.log(`Allocating net ${total_height_allocated} panelheight with base unit height ${unit_height}`)
 
     return (
         <div style={divStyle}>
@@ -28,7 +30,7 @@ const MultiTimeseriesView: FunctionComponent<Props> = ({data, width, height}) =>
                             figureDataSha1={panel.figureDataSha1}
                             isBottomPanel={ii === (data.panels.length - 1)}
                             width={width}
-                            height={H}
+                            height={Math.floor(unit_height * (panel?.relativeHeight ?? 1))}
                         />
                     </div>
                 ))
--- a/src/views/MultiTimeseries/MultiTimeseriesView.tsx
+++ b/src/views/MultiTimeseries/MultiTimeseriesView.tsx
@@ -18,7 +18,6 @@ const MultiTimeseriesView: FunctionComponent<Props> = ({data, width, height}) =>
 
     const total_height_allocated = data.panels.reduce((total, panel) => total + (panel?.relativeHeight ?? 1), 0)
     const unit_height = Math.floor(height / total_height_allocated)
-    console.log(`Allocating net ${total_height_allocated} panelheight with base unit height ${unit_height}`)
 
     return (
         <div style={divStyle}>


magland:
Looks good.

PR #33, opened 2022-02-23
Add time-span highlight support. Demo with RasterPlotView.

jsoules:
Addresses SortingView issue #182 (https://github.com/magland/sortingview/issues/182)

This PR adds functionality to handle spans of intervals that are to be highlighted in the data. As detailed in comments to the above issue, these spans are expected to be defined as a list of start times and end times connected by index, so that each span begins at `spans[0][n]` and ends at `spans[1][n]`. Arbitrarily many span-sets can be supported for any data source, and span highlight color can be left as default or set to a different color per set of spans.

One feature to call out here: if the user specifies a particular alpha, that'll be used throughout; otherwise I set the alpha to 0.5 (half-transparent) for spans that are wider than 2 pixels, and 1 (full visible) for spans smaller than that. I found that otherwise it was either too hard to see small highlight regions, or things were too opaque when zooming in.

Spans do need to be added to each time series separately. It might be possible to add an overall set of spans to a composite view, but this would require some more complex contextual logic & would come at the expense of supporting spans in individual components.

Logic to handle spans, along with defining their data types, has been incorporated as part of the `TimeSeriesView` component. It would otherwise need to be repeated across each component type, and I'd rather define it centrally.

If we like this, it can be added to any other `TimeScrollView`-based component with just a couple lines to update the data type and pass the values in.

Demo: https://www.figurl.org/f?v=http://localhost:3000/&d=3efdf652e24633ba8345fb6a02af9d01fc67e5d1&channel=flatiron1&label=SPANS_TEST

image

### DIFF
--- a/src/views/RasterPlot/TimeScrollView/TSVCursorLayer.tsx
+++ b/src/views/RasterPlot/TimeScrollView/TSVCursorLayer.tsx
@@ -1,16 +1,18 @@
 import BaseCanvas from 'FigurlCanvas/BaseCanvas';
 import React, { useMemo } from 'react';
+import { PixelHighlightSpanSet } from './TimeScrollView';
 
 export type TSVCursorLayerProps = {
     timeRange: [number, number]
     focusTimePixels?: number
+    highlightSpans?: PixelHighlightSpanSet[]
     margins: {left: number, right: number, top: number, bottom: number}
     width: number
     height: number
 }
 
 const paintCursor = (context: CanvasRenderingContext2D, props: TSVCursorLayerProps) => {
-    const {margins, focusTimePixels} = props
+    const {margins, focusTimePixels, highlightSpans} = props
     context.clearRect(0, 0, context.canvas.width, context.canvas.height)
 
     // focus time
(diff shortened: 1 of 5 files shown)


jsoules:
More zoomed in:
image

src/views/RasterPlot/TimeScrollView/TSVCursorLayer.tsx
```diff
@@ -21,13 +23,38 @@ const paintCursor = (context: CanvasRenderingContext2D, props: TSVCursorLayerPro
...
 
+// some nice purples: [161, 87, 201], or darker: [117, 56, 150]
+// dark blue: 0, 30, 255
+const defaultSpanHighlightColor = [0, 30, 255]
+
+const paintSpanHighlights = (context: CanvasRenderingContext2D, zeroHeight: number, visibleHeight: number, highlightSets: PixelHighlightSpanSet[]) => {
+    highlightSets.forEach(h => {
+        const definedColor = h.color || defaultSpanHighlightColor
+        const [r, g, b, a] = [...definedColor]
+        if (a) context.fillStyle = `rgba(${r}, ${g}, ${b}, ${a})`
```
jsoules, line 40: If a specific alpha value is given, we use that throughout.

src/views/RasterPlot/TimeScrollView/TSVCursorLayer.tsx
```diff
@@ -21,13 +23,38 @@ const paintCursor = (context: CanvasRenderingContext2D, props: TSVCursorLayerPro
...
+
+        h.pixelSpans.forEach((span) => {
+            if (!a) {
+                const alpha = span.width < 2 ? 1 : 0.5
+                context.fillStyle = `rgba(${r}, ${g}, ${b}, ${alpha})`
```
jsoules, line 45: If no alpha value was specified, use 0.5 for thick regions, or 1 for thin ones.

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
```diff
@@ -24,12 +25,23 @@ type Margins = {
...
+export type PixelSpan = {
+    start: number,
+    width: number
+}
+
+export type PixelHighlightSpanSet = {
+    pixelSpans: PixelSpan[],
+    color?: number[]
+}
+
```
jsoules, line 37: I don't know that these actually need to be exported.

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
```diff
@@ -147,12 +159,63 @@ export const use2dPanelDataToPixelMatrix = (pixelsPerSecond: number, startTimeSe
     }, [pixelsPerSecond, startTimeSec, dataMin, dataMax, userScaleFactor, panelHeight, invertY])
 }
 
+export const useTimespanToPixelMatrix = (timeToPixelMatrix: Matrix) => {
```
jsoules, line 162: Unfortunately, the actual consumer needs to get a time-to-pixels transformation matrix and then manually call this hook--we can't actually call another hook from within this class (because of the `useMemo` callback)

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
```diff
@@ -147,12 +159,63 @@ export const use2dPanelDataToPixelMatrix = (pixelsPerSecond: number, startTimeSe
...
+
+export const filterTimeRanges = (startsSec: number[], endsSec: number[], rangeStartSec: number | undefined, rangeEndSec: number | undefined): number[][] => {
+    // Take all intervals that begin before the range end or end after the range start
+    const definedRangeStartSec = rangeStartSec || 0
+
+    const startIndices = (rangeEndSec ? startsSec.map((s, i) => s <= rangeEndSec ? i : null) : startsSec.map((s, i) => i)).filter(i => i) as number[] // filter removes nulls
+    const endIndices = endsSec.map((s, i) => s >= definedRangeStartSec ? i : null).filter(i => i) as number[]
+    // get set intersection to find all those spans that begin before our window ends and end after our window begins
+    const largerSet = new Set(startIndices.length > endIndices.length ? startIndices : endIndices)
+    const smallerList = startIndices.length > endIndices.length ? endIndices : startIndices
```
jsoules, line 200: I'm not sure it actually matters which one we make the set and which one we loop over; you have to touch every index to make the hash for the set anyway.

src/views/RasterPlot/TimeScrollView/TimeScrollView.tsx
```diff
@@ -147,12 +159,63 @@ export const use2dPanelDataToPixelMatrix = (pixelsPerSecond: number, startTimeSe
...
+    const validIndices = smallerList.filter(i => largerSet.has(i))
+    const finalStarts = validIndices.map(i => Math.max(startsSec[i], definedRangeStartSec))
+    const finalEnds = validIndices.map(i => Math.min(endsSec[i], (rangeEndSec ?? endsSec[i])))
+    return [finalStarts, finalEnds]
+}
+
+export const filterAndProjectHighlightSpans = (highlightIntervals: HighlightIntervalSet[], visibleTimeStartSeconds: number | undefined, visibleTimeEndSeconds: number | undefined, spanTransform: Matrix) => {
```
jsoules, line 207: This is probably the main entry point for other components that want to use this overlay. They basically just need to provide data that came from other hooks & the actual sets of intervals that need to be highlighted.

This functionality (like the filtering functionality) is centralized here, close to where it's used.

magland:
Looks great! Merging.

PR #40, opened 2022-03-24
support ipfs uri's throughout

magland:


magland:
@jsoules this is pretty straightforward and hopefully won't conflict with anything you are working on.

src/views/Composite/CompositeViewData.ts
```diff
@@ -6,15 +6,17 @@ import { isString } from "vega"
...
     defaultHeight?: number
 }
 
 const isCVViewData = (x: any): x is CVViewData => {
     return validateObject(x, {
         label: isString,
         type: isString,
-        figureDataSha1: isSha1Hash,
+        figureDataSha1: optional(isSha1Hash), // old
+        figureDataUri: optional(isString), // new
```
jsoules, line 19: Would it be worthwhile to require that one of these two be set? (Not sure if the framework allows that, and it is checked for elsewhere in the code anyway, but it seems like we have to do that check in several places otherwise)

src/views/MountainLayout/MountainLayoutViewData.ts
```diff
@@ -6,14 +6,16 @@ import { isString } from "vega"
...
+    figureDataUri?: string // new
 }
 
 const isMLViewData = (x: any): x is MLViewData => {
     return validateObject(x, {
         label: isString,
         type: isString,
-        figureDataSha1: isSha1Hash,
+        figureDataSha1: optional(isSha1Hash), // old
+        figureDataUri: optional(isString) // new
```
jsoules, line 18: Same note as with `CompositeViewData`, is it possible to require that exactly one of these is set for the type guard to pass? I noticed it has to be checked in a couple of different consumers otherwise.

jsoules (approved):
I had a suggestion to move some of the data validation up to the type guard so that it isn't needed in the consumers, but I'm not sure if it's possible/worth it, your call.

Otherwise I don't think this'll have any impact on the stuff I've been working on, and even if it does it shouldn't be too hard to fix, so you can go ahead and merge.

src/views/Composite/CompositeViewData.ts
magland: I'm thinking of this as a transition state... eventually remove figureDataSha1 since that should really be provided as a uri starting with sha1://...

Once all the visualization plugins have been updated to do that, we can remove the old field.

PR #41, opened 2022-03-31
Restructure data element selection state with reducer.

jsoules:
Resolves SortingView issue #92 (in the sense that "gut-renovating your kitchen" resolves the issue of "leaky faucet")

Sample links:
https://www.figurl.org/f?v=http://localhost:3000&d=6326838558011a729862a3d8915f2f3e9210e990&channel=franklab2&label=J1620210531_.nwb_raw%20data%20valid%20times%20no%20premaze%20no%20home_24
https://www.figurl.org/f?v=http://localhost:3000&d=7086880aeb347d2b2a1f71e2638f7560b67e678e&channel=flatiron1&label=Jaq_03_12_visualization_data2

# Reworked data element selection state

This PR creates a contextually-shared reducer-managed state for a data set's atomic elements that would be reasonable to display in a sortable table format. In the concrete SpikeSortingView context, this means the Units displayed in the Units Table, which can be sorted by metric fields, and which need to be selectable individually or in groups of different sizes. Row metadata state (unit selection and ordering) drives UI highlights and element ordering that should be shared across several different views, and this update provides a more consistent and responsive experience.

## Motivation & Rationale

The original motivating factor was [SortingView issue #92](https://github.com/magland/sortingview/issues/92), which requested a shift-click functionality for the Units Table to select multiple elements. This functionality is surprisingly complex: to implement it correctly, the system needs to track element selection status, most-recently-clicked element, and element display order.

Updating all of these at once requires a reducer to avoid race conditions. While this was a substantial rewrite, it also offered additional benefits: selection state handling (which had previously been handled on a component-by-component basis) is now standardized, and most importantly, it enables a complete separation of row selection state and order from row contents. This leads to reduced redundant rendering and the ability to apply sorting in new places (e.g. plots).

## Outline

- `SortingSelectionContext` renamed to `RowSelectionContext` and placed in its own directory as a reducer.
  - The reducer tracks **row selection**, the **last element clicked**, **row sorting**, the list of **sorting field names**, and **pagination** (although the latter is not yet used anywhere). It also provides **callback-able base functions for element selection** click handling, in "Row Checkbox" and "Plot" flavors.
- `SortableTableWidget` and `PlotGrid` restructured to cache individual rendered elements per-row.
  - These elements are ordered and highlighted based on overall shared state, and the pre-rendered per-row contents are displayed in an explicit "[projection](https://www.ibm.com/docs/en/informix-servers/14.10?topic=concepts-selection-projection)" step.
- `SelectUnitsWidget` updated to use the `SortableTableWidget` component instead of `NiceTable`.
- Most components that use selection have been updated to refer to this advanced context reducer.

PR also includes code cleanup and restructuring changes, including moving styles into stylesheets and breaking large files into smaller components.

### Suggested reading order/Attention conservation

- I would recommend reviewing these changes starting with the [src/contexts/RowSelection tree](https://github.com/magland/spikesortingview/tree/92-units-shift-click/src/contexts/RowSelection),
- then looking at the [Units Table](https://github.com/magland/spikesortingview/blob/92-units-shift-click/src/views/UnitsTable/UnitsTableView.tsx)
- and how it is realized in the [SortableTableWidget](https://github.com/magland/spikesortingview/tree/92-units-shift-click/src/views/common/SortableTableWidget) (particularly the callbacks and row content handling).
- Then the changes for the [PlotGrid](https://github.com/magland/spikesortingview/blob/main/src/components/PlotGrid/PlotGrid.tsx) regarding callback handling, plot caching, and plot sorting.
- Changes to the [SelectUnitsWidget](https://github.com/magland/spikesortingview/tree/92-units-shift-click/src/views/common/SelectUnitsWidget) functionality are nontrivial but follow pretty clearly on the `UnitsTable` model; probably the most interesting part is how the local shadow state is handled in `useLocalSelectedUnitIds` (which maybe could be improved?)
- The rest of the changed files have fairly mechanical changes that probably aren't very interesting.

I'll also just call out that the plots displayed by the `LiveCrossCorrelationView` *do not* correspond to units individually; this is easily handled by just passing an appropriate local list.

## Specific Details

After this PR is submitted, I will provide some commentary on these specific changes in the code. The following changes are a bit tricky and benefit from explication:

- Click callback handling for rows and plots
- How sorting works in the RowSelectionContext reducer, given that the reducer has no idea what fields exist or what their contents are
- The "select-then-project" pattern used in `SortableTableWidget` (& now, `PlotGrid`) that finally works

## Testing

I have tested the following views:

In the workspace/curation view:
- Summary
- Units table
- Autocorrelograms
- Cross-Correlograms
- Average waveforms
- Electrode Geometry

I am not authorized to curate this sorting, so I haven't tested that part.

In the MultiTimeseriesView:
- Rasterplot
- Linear position
- Speed
- Position probability

Note that I'm not sure how to access unit selection in the same data set as the MultiTimeseriesView components, so I haven't tested those thoroughly, I can just confirm they still work in the context of https://www.figurl.org/f?v=http://localhost:3000&d=7086880aeb347d2b2a1f71e2638f7560b67e678e&channel=flatiron1&label=Jaq_03_12_visualization_data2 .

I would particularly note the subtle but noticeable performance improvement for select-all, particularly when there's an open plot view. Compare [this code](https://www.figurl.org/f?v=http://localhost:3000&d=6326838558011a729862a3d8915f2f3e9210e990&channel=franklab2&label=J1620210531_.nwb_raw%20data%20valid%20times%20no%20premaze%20no%20home_24) (assuming this PR's code is running locally on port 3000) and [the published version](https://www.figurl.org/f?v=gs://figurl/spikesortingview-1&d=6326838558011a729862a3d8915f2f3e9210e990&channel=franklab2&label=J1620210531_.nwb_raw%20data%20valid%20times%20no%20premaze%20no%20home_24) for J1620210531 with the units table and average waveforms displayed.

I no longer know where to find the data set referenced in https://github.com/magland/sortingview/issues/105 but I expect it would also see further improvement at this point.

### Testing

- Probably most important, I have not tested this against all components, mainly because I don't have a sense of the full scope of components available any more. I've tested against everything in the links above (the workspace and the MultiTimeseriesView), but I'm not sure what else might be out there. In particular, I'm not sure how to test that unit selection still plays nicely with the raster plot (although it should).

### Additional Features

- Should the ordering of the units in the RasterPlotView honor the sorting from the units table?

### Organization & Naming

- I've kept the `SelectionContext` part of the name for this, even though the reducer is handling a good deal more than just what rows are selected. Might want to rename.
- In the same vein, "Row" is not the most descriptive term here; the idea is that the underlying data has some sort of basic unit that is tabular, which can be given a consistent unique identifier. This should be general enough to apply to lots of scientific domains.
- All the reducer logic for the tabular state is in the `contexts/RowSelection` directory, not sure if we should have a separate directory for logic.
- I have moved many common, general-purpose views--most notably the `SortableTableWidget`, `SelectUnitsWidget`, and `BarPlot` (because I noticed it)--into `src/views/common/`. However, I'm not sure the distinction between this directory and /src/components` where e.g. `NiceTable` lives.
- Those same general-purpose views should probably be offered by the central Figurl library rather than being restricted to SpikeSortingView. (I don't remember the current state of play about shared components vs. specific application libraries.)


src/views/common/SortableTableWidget/SortableTableWidgetHeader.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const columnHeaders = useMemo(() => mapColumnsToHeaderInfo(columns, primarySortRule), [columns, primarySortRule])
+
+    const toggleSelectAllCallback = useCallback(() => selectionDispatch({type: TOGGLE_SELECT_ALL}), [selectionDispatch])
+    // We don't want the headers to have to worry about the row contents, but it makes sense for sorting logic to live with
+    // the columns. So we have the SortableTableWidget parent give us a memoized function that has the rows in it already.
+    // Then we use our own callback that changes only when that, or the columns map, changes; this one takes a set of rules
+    // and uses the wrapped function (with the rows) and the columns (which we know here) to call the sortRows function
+    // defined above. The result is a function mapping the set of rules to the ID ordering. That function is suitable for
+    // passing to the reducer.
+    const sortCallback = useCallback((rules: SortingRule[]) => rowSorterCallback(columnsMap)(rules), [rowSorterCallback, columnsMap])
```
jsoules, line 89: So above we defined a (static, not part of this component & thus not subject to re-rendering) `sorterCallbackWrapper` function that takes a set of rows and columns and returns a function mapping rules onto sorted numbers.

We also have a prop, `rowSorterCallback`, that maps columns to a function taking rules and yielding numbers. This prop is satisfied by our parent component, `SortableTableWidget`, which has a `useCallback` hook (updating only on row set changes) to wrap the table rows (with all their data fields) into the `sorterCallbackWrapper` call. That provides a function which takes the set of columns only, and yields a sorting callback (function from rules to sorted ids).

Basically, [SortableTableWidget](https://github.com/magland/spikesortingview/pull/41/files#diff-4f473aee87573e6a7158b1bedbc9d3450045e3811e5823156db47e036b896186R13) memoized the actual rowset and gave us a curried function of columns to (rules to results). Then this line memoizes & curries the column set, yielding a not-constantly-replaced function from rules to results, which we give as a callback function to the reducer in line 96.

PR #51, opened 2022-05-10
Track animation

jsoules:
This PR implements a generic framework for animations with controllers, as well as a concrete implementation for test subject position data showing the position on a test track.

### Currently Implemented:

- Generic animation framework that supports any data consisting of multiple frames
- Controls for animation framework that allow skipping, scrubbing, adjusting replay rate and direction, and non-busy-wait pausing
- Steps toward more centralized handling of native-dimension-to-pixel-dimension transform matrices
- Data types for animal position data and track data
- Working animation component that draws track and animal position

### Still to do:

- Support for streaming data (live component view) -- parts of this are present, but needs to be completed
- Support for additional position data for decoded position: this should probably be done through expanding the per-frame data type, though it could be supported by adding additional frames and layers.
- Implement other per-frame data, e.g. speed and head direction
- Actually wire up the time-step state as a Context so that other components can view it if it exists

### Interesting changes:

- in `src/FigurlCanvas/BaseCanvas.tsx`, added documentation and ability to position the canvas relative to its parent using an offset
- `src/FigurlCanvas/CanvasTransforms.tsx` offers hooks for computing layouts. These are probably better-thought-out than the overbuilt ones in `geometry.ts` and we should probably try to centralize here in the future.
- `src/views/TrackPositionAnimation/*` have a concrete implementation of animation using the animation components:
  - `TPAPositionLayer.tsx` is a canvas that draws a single frame, i.e. a dot representing animal location and a timestamp at the bottom of the graph.
  - `TPATrackLayer.tsx` is an uncomplicated canvas that just draws the track as a background frame.
  - `TrackPositionAnimationTypes.ts` describes types and type guards for track position. Note that further work will need to be done here to support streaming.
  - `TrackPositionAnimationView.tsx` is probably the key view here (after the animation reducer); it takes track data, computes margins and a transform matrix, maps native data to pixel values, gets an animation reducer, and holds the foreground and background canvases.
- `src/views/common/Animation/AnimationControl*` -- These are the parts of the generic animation module. To implement a specific animation, all the user will need to do is define a canvas with a per-frame drawing routine, and supply a reducer that tracks a frame list and a position in that frame list. Background drawing happens outside the animation module, and the framerate is assumed constant.
  - `AnimationPlaybackControls` holds the buttons (from `AnimationStateControlButtons`) that allow jump-to-beginning/end, jump-ahead/back, play/pause, reverse, and playback rate changing. These are all implemented pretty simply in the reducer. It also holds the div with the progress bar (from `AnimationStatePlaybackBarLayer`) which renders the progress bar and enables scrubbing and click-to-time-point.
  - `AnimationStateReducer` handles animation logic. Most of it is probably about what you'd expect; the tricky part is getting the tick to work right. We tick the clock using `window.requestAnimationFrame()`, but this function takes a callback whose only allowed parameter is the timestamp when the animation started to happen. So in order to allow the reducer to call its own state updates, we need to create a closure that includes a self-reference for the reducer that made it. This value is stored in the `animationDispatchFn` parameter of the state and if it is unset, animation won't work (we check for this condition). When the reducer is actually created, it then needs to be initialized with the `SET_DISPATCH` operation so that its self-reference can be populated. At that point, it has a function that can be called on every animation tick, which schedules a callback calling the `TICK` action on itself and storing the callback identifier in the state. Other state methods (e.g. `TOGGLE_PLAYBACK`) can cancel the upcoming tick using this reference; and anything that updates the state in a general sense should stop and restart the animation loop. One other trick here is that the state includes the last time that a frame was successfully rendered; since the `requestAnimationFrame()` callbacks aren't guaranteed to happen at 60 Hz, we need to check the elapsed time since the last successful render so we can keep the overall animation rate consistent. I'll comment this a little bit inline.


src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    if (s.lastRenderedTimestamp) {
+        const elapsedMs = a.now - s.lastRenderedTimestamp
+        if (elapsedMs < 0) {
+            console.warn(`Doing animation tick, but elapsed time is somehow negative: ${a.now} - ${s.lastRenderedTimestamp} = ${elapsedMs}`)
+            return s
+        }
+        const rawFramesToAdvance = elapsedMs/MS_PER_TICK * s.framesPerTick
+        const framesToAdvance = s.framesPerTick > 0 ? Math.floor(rawFramesToAdvance) : Math.ceil(rawFramesToAdvance)
```
jsoules, line 185: Of course, if we're going backwards, we need to CEIL rather than FLOOR or we animate too fast.

src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -0,0 +1,231 @@
...
+        s.currentFrameIndex = Math.min((s.frameData.length - 1), s.currentFrameIndex + framesToAdvance)
+        if (framesToAdvance !== 0) {
+            s.lastRenderedTimestamp = a.now
+        }
+    } else {
+        s.lastRenderedTimestamp = a.now
```
jsoules, line 191: This closes the initial check to see if `lastRenderedTimestamp` was defined. If it was undef, we're coming off a pause, and will start actually animating on the next cycle.

src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    if (((s.currentFrameIndex === s.frameData.length - 1) && s.framesPerTick > 0) || ((s.currentFrameIndex === 0) && s.framesPerTick < 0)) {
+        return togglePlayState(s)
+    }
```
jsoules, line 196: This automatically pauses when we reach the end of the recording (in the direction we're going).

magland:
@jsoules this is amazing!
The generic component could be super-useful for many other views, so I'm glad you did it this way.
High quality rendering and very responsive.
Controls are very intuitive.
I glanced through the code and it all looks reasonable.

Here's the live link:
https://v1.figurl.org/f?v=gs://figurl/spikesortingview-1&d=0571007f92dce1ef7d4d1d8594fd70880332cd40&channel=flatiron1&label=Test%20workspace

I'll open a couple of issues for some minor improvements, but I'll merge for now.

magland:
@jsoules
I have merged this functionality into the figurl2 branch, and so here's the (preferred) link that uses ipfs/kachery-cloud:

https://www.figurl.org/f?v=gs://figurl/spikesortingview-2&d=ipfs://bafybeiaee44ndl4w3gvsvg32ctimb62ilbfuh5y3siu4nhrozv2bo6rmky&label=Test%20workspace

Let's do further development on figurl2 branch (until of course it is merged into main)

Issue #52, opened 2022-05-10
playback rate suggestion


magland:
Rather than having up/down controls, I think it would be more intuitive and more useful to have a dropdown selection (or maybe it would need to be a "dropup" selection) that allows user to select between a discrete set of choices:
x0.25 x0.5 x1 x2 x3 x4 x5 x10

Here's the existing control:
![image](https://user-images.githubusercontent.com/3679296/167691951-7ba63f24-1d42-4f06-b094-5cb64658d080.png)

PR #55, opened 2022-05-11
Implement replay-rate dropdown menu selection.

jsoules:
Fixes issue #52.

Example:
image

Used `react-select` library with some recommended modifications for inline styling to allow user to choose from a palette of replay rates instead of using the button-based stepwise adjustment. An individual component can select whether to present the button-based or dropdown-based interface based on a toggle prop.

By default the replay rate selection presents values of 0.1x, 0.25x, 0.5x, 1x, 2x, 3x, 4x, 5x, and 10x, but additional replay rates can be passed through the `customPlaybackRates` parameter; the combined list will be sorted and uniqued before being presented to the user. There is not currently any functionality to disable the default rate values; this didn't seem necessary but would be easy enough to implement if desired. Additionally, the dropdown options don't currently show negative signs while playing in reverse; this would be trivial to implement if desired.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "react-draggable": "^4.4.4",
     "react-icons": "^4.3.1",
     "react-scripts": "4.0.3",
+    "react-select": "^5.3.2",
     "react-sizeme": "^3.0.2",
     "react-vega": "^7.4.4",
     "typescript": "^4.1.2",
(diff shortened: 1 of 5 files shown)


magland:
@jsoules I switched the base of this PR to figurl2 -- I think that's what you intended.

Seems to work perfectly.

jsoules:
Yep--thanks! I had just realized I did the wrong base, glad you caught that!

PR #56, opened 2022-05-18
Add head direction

jsoules:
Not ready for merge yet--needs more evaluation. See link:
https://www.figurl.org/f?v=http://localhost:3000&d=ipfs://bafybeibantoudovpweoef2uxsm5exdq3lfrooqwix4htoxc225hrsuur6q&label=example%20track%20animation

I wonder if there's some kind of rotation here between the head and the way the track is plotting out? Or does the animal just walk backwards a lot?


magland:
@jsoules, should we go ahead and merge? Seems to work very nicely.

jsoules:
Sure, we should be able to merge this. I don't think I had anything outstanding on it, provided the lab thinks the orientation looks as expected.

PR #58, opened 2022-05-25
Add base replay rate for 1:1 playback.

jsoules:
For review--we should discuss before merging. Note merge target is the head direction branch.

This PR adds a real-time (1:1) replay rate figure in MS, and keys the other replay rate timings off this multiplier (which defaults to 16.6-bar or 60 fps).

Example: https://www.figurl.org/f?v=http://localhost:3000&d=ipfs://bafybeiaanv2dz5gazhynganbeibd7xnkefrrcjwxetamrmcylfe3wndyju&label=example%20track%20animation

In the load script (different project so not part of this PR), we compute and include the `realTimeReplayRateMs` as follows:

```python
# timestamps = 1-dimensional numpy array of the times
vals, counts = np.unique(np.diff(timestamps), return_counts=True)
idx = np.argmax(counts)
interframe_ms = vals[idx] * 1000
data['realTimeReplayRateMs'] = interframe_ms
```
which sets the elapsed time per frame to the modal value of the observed differences between frame times in the data set.

The front-end in this PR has some debug code to report the base MS per frame (which should be removed before it is accepted) whenever the playback is paused. This confirms that the base MS per frame in the example is the expected value of 33.619-something ms.

However, note that playback is not actually real-time at this rate: empirically (timing for 10 elapsed seconds per the displayed timing in the recording) 10 seconds of playback takes about 15 seconds of real time. I expect this is due to accumulated rounding error/incompatibility between the 60-fps refresh rate for the animation callbacks and the desired frame rate--since the "time last frame was drawn" has to get reset every cycle, instead of having a frame rate seesaw, instead we just have consistent lag behind the intended frame rate so that we actually realize something like 20 FPS instead of the 29.75 intended FPS.

### DIFF
--- a/src/views/TrackPositionAnimation/TrackPositionAnimationTypes.ts
+++ b/src/views/TrackPositionAnimation/TrackPositionAnimationTypes.ts
@@ -35,7 +35,7 @@ export type PositionFrame = {
  * @member ymin Lowest y-value to display, in native units.
  * @member ymax Highest y-value to display, in native units.
  * @member headDirection Direction of the subject's head in the xy-plane, in radians.
- * @member realTimeReplayRate Optional, assumed 1. If set, specifies the initial frames-per-tick for playback.
+ * @member realTimeReplayRate Optional, assumed 1000/60. If set, specifies the number of milliseconds per frame to achieve real-time playback.
  */
 export type TrackAnimationStaticData = {
     type: 'TrackAnimation'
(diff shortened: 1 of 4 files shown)


magland:
Seems that realTimeReplayRateMs is the  number of milliseconds per frame. So it's not really a "rate". The rate would be the reciprocal of this.

I don't think "realTime" should be part of this name either. Really what we're talking about is the sampling frequency (so maybe samplingFrequency in Hz)

I think this should be computed on the javascript side rather than in python, since it can be determined from the timestamps array. That is sent to the browser, correct?

I think there are ways to get the speed to be accurate relative to wall clock. We can discuss.

jsoules:
I just pushed a change that reworks how the framework handles animation book-keeping; it looks like wall-clock times are now more accurate.

`realTimeReplayRateMs` is the number of milliseconds (per frame) that achieves a real-time replay rate. :) But we can call it something else.

The frame spacing can be computed by the javascript, but that will require it to be recomputed for every run of the animation (which in turn could involve looking at the entire timestamp dataset to find the modal amount, if we don't want to just project from the first couple frames)--numpy is probably better for this than math.js, and in any event it seems worthwhile to do the computation once and be done.

magland:
Merging this and closing the head position PR.

PR #59, opened 2022-05-27
WIP: implement linear position overlay 

magland:
A bit hacky, but this allows linear position overlay on the pdf position plot.


jsoules:
I know this is still WIP, just wanted to confirm that I've had a look at it.
Only thought is whether it'd be worth modifying the alpha channel for the overlay so that it's semi-transparent--I wonder if that might help address some of Eric's concerns about obscuring the predictions.

In the longer run, we might want to try to modify the TimeScrollView component to allow it to have multiple content panes, and refactor the content pane modules so that they can be composed over a TimeScrollView--basically allowing an arbitrary number of TSV-based components to be layered on top of each other while using the same underlying logic they'd use as individual components--but that's not necessary for the immediate ask; it'd probably be a good feature for the overall Figurl framework but isn't high-priority unless Franklab has multiple other examples of timeseries they want to stacked on the same plot.

edeno:
@magland I see this PR for the linear position overlay, but I haven't been able to find any of the python code used to generate the overlay. Is there somewhere else I should look? Apologies if I'm not finding it.

PR #61, opened 2022-07-25
Sync focus time with animation

jsoules:
This PR includes some logic and links to make the `TrackPositionAnimation` aware of the context-global focus time from the `RecordingSelectionContext` object.

The implemented behavior is as follows:

- A click in a view that sets the focus time (e.g. the raster plot view) will update the focus time. The new focus time will be communicated to the animation, which finds the frame with the closest time code and sets that as the currently active frame. Playback (if in progress) will continue uninterrupted from that point.
- The time represented by the currently active playback frame will be set as the focus time in the context whenever playback starts or stops. This allows other views to sync to the currently visible animation frame.

In a future version, I will probably alter the condition for syncing from the animation to the other views, so that they can scroll in real-time--this will probably involve logic in the default `TimeScrollView` to recenter the recording viewing window if the focus time is set outside the visible window, but that needs more feature design before we can implement it.

### DIFF
--- a/src/util/bst.ts
+++ b/src/util/bst.ts
@@ -2,52 +2,103 @@ import { useMemo } from "react"
 
 export type BstNode<T> = {
     value: T | undefined
+    baseListIndex: number | undefined
     left: BstNode<T> | undefined
     right: BstNode<T> | undefined
 }
 
-const makeBinaryTree = <T,>(items: T[]): BstNode<T> => {
+export interface ValueWithPosition<T> {
+    value: T,
+    baseListIndex: number
+}
+
+export interface BstSearchResult<T> extends ValueWithPosition<T> {
+    offBy: number
+}
+
+export type BstSearchFn<T> = (value: T, requireExact?: boolean) => BstSearchResult<T> | undefined
+export type MetricFn<T> = (a: T, b: T) => number
+
+export const makeBinaryTree = <T,>(items: ValueWithPosition<T>[]): BstNode<T> => {
     switch (items.length) {
         case 0:
             console.log(`Encountered empty array in building binary search tree: this probably shoudn't happen`)
-            return { value: undefined, left: undefined, right: undefined } as BstNode<T>
+            return { value: undefined, left: undefined, right: undefined, baseListIndex: undefined } as BstNode<T>
         case 1:
-            return { value: items[0], left: undefined, right: undefined } as BstNode<T>
+            return { value: items[0].value, left: undefined, right: undefined, baseListIndex: items[0].baseListIndex } as BstNode<T>
         case 2: // the node will have only 1 child! Let's arbitrarily make it the left one
-            return { value: items[1], left: makeBinaryTree([items[0]]), right: undefined } as BstNode<T>
+            return { value: items[1].value, left: makeBinaryTree([items[0]]), right: undefined, baseListIndex: items[1].baseListIndex } as BstNode<T>
         default:
             const split = Math.floor(items.length / 2)
             const leftChild = makeBinaryTree(items.slice(0, split))
-            const value = items[split]
+            const thisNode = items[split]
             const rightChild = makeBinaryTree(items.slice(split + 1))
-            return { value: value, left: leftChild, right: rightChild } as BstNode<T>
+            return { value: thisNode.value, baseListIndex: thisNode.baseListIndex, left: leftChild, right: rightChild } as BstNode<T>
     }
 }
 
 
-export const useBinarySearchTree = <T,>(items: T[], needsSort: boolean = false, compFn: (a: T, b: T) => number): BstNode<T> => {
+const nodeToValue = <T,>(node: BstNode<T>, difference: number): BstSearchResult<T> | undefined => {
+    if (!node || node.value === undefined || node.baseListIndex === undefined) return undefined
+    return {
+        value: node.value, baseListIndex: node.baseListIndex, offBy: difference
+    }
+}
+
+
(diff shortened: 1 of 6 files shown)


jsoules:
See https://www.figurl.org/f?v=gs://figurl/spikesortingview-6dev3&d=sha1://ff184ea76343c6fb036fdd0ac75154868317567d&label=Time%20click%20test for a working example--this joins a test raster plot (with a SpikeInterface toy example) with the non-decoded recording animation.

src/util/bst.ts
```diff
@@ -0,0 +1,107 @@
...
+        const sorted = labeledItems.every((item, index, array) => {
+            return index === 0 || metricFn(item.value, array[index - 1].value) >= 0
+        })
+        if (!sorted) {
+            labeledItems.sort((a, b) => metricFn(a.value, b.value))
+        }
+
+        const root = makeBinaryTree(labeledItems)
+        return getSearchFn(metricFn, root)
+    }, [items, metricFn])
```
jsoules, line 104: Here we attempt to speed up matching the click with the nearest frame by building a binary search tree of the time series. This code could be used for any static sorted list. As the note in the code says, we might not actually need this much infrastructure--having a sorted list, we could just use a binary search algorithm on the list itself--so it might be easier to just do away with this class.

src/views/TrackPositionAnimation/TPATimeSyncLogic.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useCallback, useMemo } from "react"
+import useBinarySearchTree, { BstSearchFn, BstSearchResult } from "util/bst"
+import { AnimationState, AnimationStateAction } from "views/common/Animation/AnimationStateReducer"
+import { PositionFrame } from "./TrackPositionAnimationTypes"
```
jsoules, line 4: This file provides three basic bits of functionality:
- Given the time-series data, it constructs a memoized binary search tree of the times & the corresponding frames, and returns a search function generator that makes it easy to find the frame closest to any given time code
- A `matchFocusToFrame` function that updates the focus time given the animation state (which identifies the current frame) if the current frame's time is more than a minimum distance from the focus time. This is expected to be called when the animation playback mode changes (but could handle other calling conditions too)
- A `matchFrameToFocus` function that should be called in a `useEffect` hook driven by the focus time; this sets the animation playback current frame to the one closest to the focus time when the focus time changes. (The hook should be set up so this triggers only when the change is external to the track position animation view.)

src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
```diff
@@ -164,6 +167,13 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
         })
     }, [dataFrames])
 
+    // TODO: Full support of dynamic updates to frame set will require looking at AnimationState.frameData instead of dataFrames!
+
+    const { focusTime, setTimeFocus } = useTimeFocus()  // state imported from recording context
+    const findNearestTime = useTimeLookupFn(animationState)
+    useEffect(() => matchFrameToFocus(focusTime, findNearestTime, animationState, animationStateDispatch), [focusTime])
+    useEffect(() => matchFocusToFrame(animationState, focusTime, setTimeFocus), [animationState.isPlaying])
+
```
jsoules, line 176: Here's the time update code in action. The nearest-time-lookup function is provided by a custom hook that depends on the animation state's frame data (and thus should update correctly if, in the future, we have a dynamically loaded time series). The two `useEffect` hooks should control the triggers and direction of the time updates: match the frame to the focus if the focus changes; and match the focus to the frame if we change playback state.

src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -137,36 +137,43 @@ const AnimationStateReducer = <T, >(s: AnimationState<T>, a: AnimationStateActio
...
             return {...s, playbackStartedTimestamp: undefined}
         default: {
             throw Error(`Invalid action type for animation state reducer: ${type}`)
         }
     }
 }
 
+// It's not actually clear that this ever did anything useful.
+// TODO: Deprecated. Remove this function (& references to it) after Sep 2022
+// if the widget's behavior remains normal.
```
jsoules, line 164: This function doesn't appear to have actually done anything useful, but I think it was also responsible for a race condition when multiple chains of animation-update requests could get triggered on changing the current frame. I've commented it out for now to see if anything breaks that I didn't notice during testing, but it should likely be removed at some point in the near future.

src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -187,6 +194,13 @@ const doTick = <T, >(s: AnimationState<T>, a: AnimationStateTickAction): Animati
         console.warn('Attempt to animate, but no dispatch function set. No-op.')
         return s
     }
+    if (!s.isPlaying) {
+        // Shouldn't happen, but these sometimes slip through on rapid interactions.
+        console.log(`Caught unexpected tick while theoretically paused.`)
+        s.pendingFrameCode && window.cancelAnimationFrame(s.pendingFrameCode)
+        s.pendingFrameCode = undefined
+        return {...s}
+    }
```
jsoules, line 203: This also manually corrects for the race condition by ensuring the animation callback chain will stop itself if the state toggles to not-playing. (Previously this was handled only by the code that toggled playback state.) It's still possible to trip this up by clicking rapidly on the play/pause button, so we might want to do some more formal debouncing, but this seems to work pretty well now.

jsoules:
This PR's content was incorporated into PR #63, which has now been accepted, so this is no longer meaningful.

PR #63, opened 2022-08-09
Animation windows

jsoules:
This PR comprises two main parts. (Note that it is forked off of the prior PR #61 so if this is accepted, that one will no longer be needed.)

The first is a lot of code cleanup and breaking apart components into smaller, hopefully more reusable and easier-to-follow files. I've tried to centralize as much as possible of the reusable animation/animation control elements into the `views/common/Animation` directory.

The second part is code to add some new windowing functionality to the animation playback. The prior PR includes synchronization of the focus time (for a static timeseries graph such as a raster plot) with the time associated with a frame of animation. This PR expands on this by adding a sync button (to restrict the visible part of the playback scrub bar to only the window shown in the overall zoomed state of other time series) and the ability to drag-select and restrict to a specific subsection of the overall animation. I've also implemented (debounced, so still performant) live syncing of the focus bar, so that the focused time point in a time series chart will move in real-time along with the playback of an animation.

From a technical side, other interesting contributions are a standardized throttle/debounce architecture, which all the animation components now use (in `util/rateLimiters.tsx`) as well as generic syncing for animations when their frame data contains a time code (code located in `views/common/Animation/AnimationUtilities` and used for the `PositionFrame` case in the `TrackPositionAnimation` element). I've improved the animation widget to handle both left-side (main) controls and right-side (optional, expandable) control buttons, with the scrubbing bar adjusting between them. I've also updated the `Animation` widget design so that controls are standardized and don't have to be instantiated by a consuming element--layers for an animation frame are now passed to the `FrameAnimation` element as its children, and get displayed in the space that isn't taken up by the scrub bar and controls.


src/util/rateLimiters.tsx
```diff
@@ -0,0 +1,66 @@
+import { useCallback, useMemo, useRef } from 'react'
```
jsoules, line 1: This file defines two hooks for closely related rate limiters which we've wound up using a lot.
A *debouncer* collects a user input stream and holds off on processing it until it has been quiet for a particular period of time (e.g. type-ahead search not rerunning the search until the user has stopped typing for 200 ms).
A *throttler* keeps a function from updating more than once every x (while still changing the ultimate effect of the update up until the time it's run).

These are essentially the same rate-limiting function, the question is just do we cancel and reschedule while input is incoming (debouncer) or do we leave the scheduled update, but not schedule a new one until that one has cleared (throttler).

Anyway, I've decomposed both of these types into a couple pieces. From my perspective, both consist of an *updater* function (which tracks the new target state and handles scheduling of its resolution) and a *resolver* function (which actually updates the permanent state), as well as a set of *refs* (which can be modified rapidly without triggering rerenders or updates).

The general flow is that the refs are created and persisted, the updater changes the values in the refs and schedules resolution if the update makes the state different from the persistent state, and the resolver (when it finally gets to run) makes the persisted state look like what's in the refs.

Probably a good practical example is in the `views/common/Animation/AnimationControls/useWheelHandler.tsx` file, which handles throttled updates to the current playback frame as the user scrolls through with the mouse wheel. We only update the actual current frame a maximum of 60 times per second, no matter how fast you move the scroll wheel; but we do collect a count of the number of scrolls in any particular direction in order to determine how much to adjust the frame state.

src/views/TrackPositionAnimation/TPADecodedPositionLayer.tsx
```diff
@@ -39,8 +39,8 @@ const stylesMap = {
     'base': valuesRange.map((i) => `rgba(${baseRed}, ${baseBlue}, ${baseGreen}, ${i/5})`)
 }
 
-const draw = (context: CanvasRenderingContext2D, props: DecodeFrameProps) => {
-    const { frame, colorMap } = props
+const draw = (context: CanvasRenderingContext2D, props: DecodeFrameProps, colorMap?: ValidColorMap) => {
```
jsoules, line 42: This is that "callbacks-as-configuration" stuff I was talking about: instead of having the color map passed to the draw function as part of the frame data (props) for every frame, we configure the draw function with the callback hook in l. 64-66.

src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
```diff
@@ -90,24 +93,6 @@ const usePositionFrames = (positions: number[][] | undefined, timestampStart: nu
     }, [positions, timestampStart, timestamps, headDirection, decodedData])
 }
 
-const setupAnimationStateDispatchFn = (animationStateDispatch: React.Dispatch<AnimationStateAction<PositionFrame>>) => {
```
jsoules, line 93: This stuff is relevant to every `FrameAnimation` and so has been moved to the component itself.

src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
```diff
@@ -155,6 +137,7 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
     }, [data.decodedData])
     const dataFrames = useFrames(data.positions, decodedData, transform, headDirection, data.timestampStart, data.timestamps)
     const decodedLocationsMap = useProbabilityLocationsMap(transform, decodedData)
+    useRecordingSelectionTimeInitialization(data.timestamps[0] + (data.timestampStart ?? 0), data.timestamps[data.timestamps.length - 1] + (data.timestampStart ?? 0))
```
jsoules, line 140: Maybe I should move this out to the rest of the time-handling stuff.

src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
```diff
@@ -164,14 +147,22 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
...
+
+    const { handleOutsideTimeUpdate, handleFrameTimeUpdate } = useLiveTimeSyncing(setTimeFocus, animationState, animationStateDispatch, getTimeFromFrame)
+    useEffect(() => handleOutsideTimeUpdate(focusTime), [handleOutsideTimeUpdate, focusTime])
+    useEffect(() => handleFrameTimeUpdate(), [handleFrameTimeUpdate])
+
+    useTimeWindowSyncing(animationState, animationStateDispatch, getTimeFromFrame)
+
     const currentProbabilityFrame = useMemo(() => {
-        const linearFrame = dataFrames[animationState.currentFrameIndex].decodedPositionFrame
+        const linearFrame = animationState.frameData[animationState.currentFrameIndex]?.decodedPositionFrame
```
jsoules, line 160: `dataFrames` comes from the input data, while `animationState.frameData` is the complete data set that's currently persisted in the state.

If we ever manage to get buffered loading (or other ways of incrementally building the state frames) set up, the distinction will be important.

src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
```diff
@@ -185,39 +176,34 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
...
-            <div>
-                {trackLayer}
-                {probabilityLayer}
-                {positionLayer}
-                {controlLayer}
-            </div>
-        </Fragment>
+            dataSeriesFrameRateHz={samplingFrequencyHz}
+        >
+            {trackLayer}
```
jsoules, line 203: Here the content layers are passed as children of the `FrameAnimation` component, which is responsible for the control bars. No need for future animations to duplicate control-bar code.

src/views/common/Animation/AnimationControlButtonContainer.tsx
```diff
@@ -0,0 +1,38 @@
+import { FunctionComponent, ReactNode, useMemo } from 'react'
```
jsoules, line 1: Buttons have been separated into their own files; the div that contains them is this `AnimationControlButtonContainer` and handles margins and spacing between them and stuff.

Basically I didn't want to repeat the code between the left and right control panels.

src/views/common/Animation/AnimationControlButtonStyles.css
```diff
@@ -1,6 +1,12 @@
 div.AnimationControlButton span {
-    padding-left: 4px;
-    padding-right: 4px;
+    padding: 4px;
+    /* text-align: center; /* Ought to work, but doesn't, alas. */
```
jsoules, line 3: For some reason this styling isn't honored properly when I put it in the CSS instead of in explicit properties.

src/views/common/Animation/AnimationControls/PlaybackOptionalButtons.ts
```diff
@@ -0,0 +1,8 @@
+
```
jsoules, line 1: This file collects the set of buttons which can optionally be included (and whose presence/absence will dynamically change the width of the right-hand control panel).

src/views/common/Animation/AnimationControls/PlaybackChangeDirectionButton.tsx
```diff
@@ -0,0 +1,41 @@
+import { faStepBackward, faStepForward } from '@fortawesome/free-solid-svg-icons'
```
jsoules, line 1: Breaking these buttons into their own components makes them easier to manage and makes the other controls way more lightweight.

src/views/common/Animation/AnimationControls/useDraggableScrubber.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { useCallback, useMemo, useRef } from 'react'
```
jsoules, line 1: This guy uses the throttler from the rateLimiters, while also exposing some logic that determines whether and how to handle initiating and concluding the overall scrub action.

src/views/common/Animation/AnimationControls/usePlaybackBarGeometry.tsx
```diff
@@ -0,0 +1,81 @@
+import { norm } from 'mathjs'
```
jsoules, line 1: This file breaks out (and memoizes) most of the stuff that describes the scrub bar based purely on its geometry. The idea is a) to make that more straightforward and b) to avoid rerunning the parts of hooks that don't need to be rerun just because the current time or displayed-frame changed--we don't need to recompute the padding on the bars and all that just because we changed what frame is up.

src/views/common/Animation/AnimationControls/usePlaybackBarGeometry.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const framesPerPixel = useMemo(() => (visibleWindow[1] - visibleWindow[0])/barWidth, [visibleWindow, barWidth])
+
+    const xToFrame = useCallback((x: number) => _xToFrame(x, framesPerPixel, visibleWindow[0], barWidth), [framesPerPixel, visibleWindow, barWidth])
+    const getEventPoint = useCallback((e: React.MouseEvent) => _getEventPoint(e, (leftOffset + _styling.leftMargin)), [leftOffset, _styling.leftMargin])
+    const frameToPixelX = useCallback((elapsedFrames: number) => _frameToPixelX(elapsedFrames, framesPerPixel), [framesPerPixel])
+
+    const getBarInterpreter: LogicalBarInterpreterGenerator = useCallback(
+        (scrubberCenterX) =>
+            (x: number, y: number) => _barInterpreter(x, y, scrubberCenterX, _styling.scrubberRadius, barCanvasVCenter),
+        [_styling.scrubberRadius, barCanvasVCenter])
```
jsoules, line 72: Unfortunately, one complication is that we do actually need to know the geometry and the current frame (and overall frame count) in order to interpret whether the user clicked on the scrubber or off it. Thus, this callback generates a click interpreter when given the scrubber's pixel position (which depends on the current frame) but doesn't need to recompute the overall bar dimensions etc.

src/views/common/Animation/AnimationControls/usePlaybackBarPixelState.ts
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react'
```
jsoules, line 1: And this file handles the playback bar configuration that actually changes based on the current frame index, elapsed frames, overall frame count, etc.

src/views/common/Animation/AnimationControls/usePlaybackBarSelection.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useCallback, useMemo, useRef } from 'react'
```
jsoules, line 1: Implements the ability to drag-select parts of the scrub bar in order to create a sub-window, basically zooming in on a particular section of the animation.

src/views/common/Animation/AnimationControls/useWheelHandler.tsx
```diff
@@ -0,0 +1,64 @@
...
+    refs.useFineWheel.current = false
+    refs.wheelCount.current = 0
+}
+
+export const useThrottledWheel = (dispatch: AnimationStateDispatcher<any>) => {
+    const wheelCount = useRef<number>(0)
+    const useFineWheel = useRef<boolean>(false)
+    const refs = useMemo(() => {return { wheelCount, useFineWheel }}, [wheelCount, useFineWheel])
+    const resolverProps = useMemo(() => { return { dispatch }}, [dispatch])
+    const wheelHandler = useThrottler(wheelUpdate, wheelResolver, refs, resolverProps)
```
jsoules, line 48: The general pattern is that the throttler maintains its own internal state and handles calling of the resolver: it returns to the consumer an interface which is just a function that takes only the values that might need to be updated. The refs that store them and the resolver that persists them to the actual state have both been away.

src/views/common/Animation/AnimationStatePlaybackBarLayer.tsx
```diff
@@ -1,175 +1,156 @@
...
-            dispatch({type: 'SET_CURRENT_FRAME', newIndex: newFrame})
-        }
-    }, [getEventPoint, drawData.seenX, barAreaVCenter, isPlaying, dispatch, getNewFrame])
-
+        if (initiateScrubbing(x, y, isPlaying)) return // if the user grabbed the scrubber, don't do anything else
+        dragSelectInitiationRef.current = window.setTimeout(() => {
+            dragSelectInitiationRef.current = undefined
+            initiateDragSelection(x)
+        }, dragSelectInitiationDelayMs)
+    }, [getEventPoint, isPlaying, initiateScrubbing, initiateDragSelection])
```
jsoules, line 110: So the trick with drag-select and also supporting click-selection is that you don't know from a click whether you are starting a drag or just clicking.

To get around this, the way this is handled is when the user clicks, this schedules a callback. If the callback resolves, the state is set to "dragging-selecting." If a mouseup event happens before the callback resolves, the mouseup cancels the callback and treats it as a click interaction (which moves to the clicked-on frame).

src/views/common/Animation/AnimationStatePlaybackBarLayer.tsx
```diff
@@ -1,175 +1,156 @@
...
-            if (nextDebounceRequest.current) {
-                window.cancelAnimationFrame(nextDebounceRequest.current)
-            }
-            nextDebounceRequest.current = undefined
-            dragTargetFrameRef.current = undefined
-        }
-    }, [currentFrameIndex, dispatch])
-
+        terminateScrubbing(isPlaying)
+        terminateDragSelection()
```
jsoules, line 124: These guys should be no-ops if the underlying state isn't relevant (i.e. we aren't scrubbing and/or dragging).

src/views/common/Animation/AnimationStatePlaybackBarLayer.tsx
```diff
@@ -1,175 +1,156 @@
...
-
-        const [x, y] = getEventPoint(e)
-        if (Math.abs(x - draggingXRef.current) > 2) {
-            dragTargetFrameRef.current = getNewFrame(x, y)
-            draggingXRef.current = x
-            debounceSetFrame()
+        if (e.buttons === 0) { // terminate drags where mouseup happened outside our element
+            terminateScrubbing(isPlaying)
+            terminateDragSelection()
+            return // short-circuit processing of no-button mouse moves--we don't care about them
```
jsoules, line 131: This can happen if the user clicks, drags, then releases the mouse button outside the frame. Then no mouseup event ever gets generated, so we keep scrubbing/dragging even though the button isn't held down.

src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -42,6 +53,38 @@ export type AnimationStateUpdateFrameDataAction<T> = {
...
+}
+
+export type AnimationStateCommitWindowAction = {
+    type: 'COMMIT_WINDOW'
+}
+
+export type AnimationStateReleaseWindowAction = {
+    type: 'RELEASE_WINDOW'
+}
+
```
jsoules, line 83: The overall animation state tracks the (proposed or realized) visible window. This is technically necessary (since we can't just persist it as pixels--the data flow doesn't work out right) but also means that you can bounce back and forth from the restricted view to the full-length view without losing the window selection.

src/views/common/Animation/AnimationUtilities/useTimeWindowSyncing.ts
```diff
@@ -0,0 +1,29 @@
+import { useTimeRange } from 'contexts/RecordingSelectionContext'
```
jsoules, line 1: I think this was in the previous PR.

src/views/common/Animation/AnimationUtilities/useTimeSyncing.ts
```diff
@@ -0,0 +1,122 @@
+import React, { useCallback, useMemo, useRef } from "react"
```
jsoules, line 1: This is currently used with the focus time from the recording, but it could be used for any frame that contains time data and external time source/sink. The question is which way to run the updates.

src/views/common/Animation/AnimationUtilities/useTimeSyncing.ts
```diff
@@ -0,0 +1,122 @@
...
+
+type liveSyncFrameToOutsideTimeProps = {
+    outsideTime: number | undefined
+    checkNewTimeWithinEpsilon: (focusTime: number | undefined, replayRateMultiplier: number) => boolean
+    multiplier: number
+    cancelPendingUpdateOfOutsideTime: () => void
+    findNearestTime: (time: number) => BstSearchResult<number> | undefined
+    dispatch: React.Dispatch<AnimationStateAction<any>>
+}
+export const liveSyncFrameToOutsideTime = (props: liveSyncFrameToOutsideTimeProps) => {
```
jsoules, line 62: This syncs the frame (i.e. the playback time code) to the time from the external source (i.e. the focus time from the recording context), IFF those times are more than some epsilon value apart.

If you don't do this, you get an infinite loop: the focus time will update to match the playback time, but that's outdated by the time the callback fires.

The epsilon value is in turn set by using 1.5 times the playback rate times the throttling rate: if we run the update once every 100 ms, we expect the new time to be within 100 ms of the old focus time. Only I use 1.5 times that, since things slip. And if we're playing at 10x speed, then it'd be 1.5 s instead of 150 ms. etc.

src/views/common/Animation/AnimationUtilities/useTimeSyncing.ts
```diff
@@ -0,0 +1,122 @@
...
+        findNearestTime,
+        dispatch
+    }), [epsilonCheck, state.replayMultiplier, cancelThrottled, findNearestTime, dispatch])
+
+    const handleFrameTimeUpdate = useCallback(() => {
+        const internalTime = getTimeFromFrame(state?.frameData[state?.currentFrameIndex])
+        state.isPlaying ? throttledSyncOutsideTimeToFrameTime({currentTime: internalTime}) : syncOutsideTimeToFrameTime(internalTime, setExternalTime)
+    }, [getTimeFromFrame, state.isPlaying, state.frameData, state.currentFrameIndex, throttledSyncOutsideTimeToFrameTime, setExternalTime])
+
+    return { handleOutsideTimeUpdate, handleFrameTimeUpdate }
```
jsoules, line 115: At the end of the day, all of this logic exposes two functions which should be called in `useEffect` hooks in the consumer. One of them should be called every time the external time (i.e. focusTime) changes: that's the one that syncs playback frame to the external time frame. During playback this will do nothing if the focus time and the playback time are close enough--in order to account for the throttled callback updating the focus time from the playback time--but clicking to update the focus still works during playback for realistically far-away values.

The other function can be run whenever playback is started or stopped, and just syncs the focus state to the frame state.

src/views/common/Animation/AnimationUtilities/useTimeSyncing.ts
jsoules: Sorry, to expand on that thought a little:

This is related to ensuring that automatic scrolling of the focus time (in the raster plot) works during active playback of the animation, while still letting a click in the raster plot (or on the scrub bar) reset the time.

There are two times: there's the outside time (here the focus time in the recording context--the one you could set by e.g. clicking the raster plot) and the time that's endogenous to the animation in active playback (call it the "frame time" as it's the time associated with the frame currently being displayed, which is updating ~60 times a second). The tricky part is that neither one is the source of truth: we want the selected frame to update if the user clicks to change the focus, but we also want the animation (during playback) to update the focus time to match the current frame.

This can produce a cycle: the frame animates to the next frame, which updates the frame time, which tries to update the focus time, which triggers an update that resets the frame time. It's all due to the hook that resets the current frame to match the focus time--that hook can only fire when the focus time updates, but there's nothing to distinguish a focus time change that should be honored (due to raster plot click) vs. a focus time that's a lagging effect of the animation updating the focus time.

We got around this in the past by turning off "live" syncing of the focus time: the hook that updates the focus time from the frame time was skipped entirely if playback was active. That way, a manual click (to change the focus time) always works and will update the focus time, triggering an update of the frame time that overwrites the next-frame call; and we just update the focus whenever playback stops (triggering a single focus-time update that won't interfere with the no-longer-ongoing animation of the frames). But this doesn't allow live movement of the focus line in the raster plot.

This only gets worse when you have throttling: the animation will play at one speed, the focus time updates are throttled (for performance reasons) to a lower speed; and as a result the updates to the focus will never quite sync up with the time implied by the current frame, and you can get stuck in further cycles or introduce playback lag.

The way I've implemented it here is to introduce an "epsilon" distance which measures the difference between the current focus time and the time the animation playback wants to update the focus to. Basically, we still have a "reset the frame to the focus" hook that fires on every update of the focus--EXCEPT that hook now knows the time difference between the current focus state and the updated focus state. It also knows how fast playback is happening (based on the multiplier). So if the new focus time is within the playback-rate-adjusted timespan implied by the sync throttling interval, the hook knows not to try to update the frame to match the focus, and just accepts the new focus time. Instead of deferring updates to the focus time until playback has stopped, we just break the potentially cyclic updates by conditionally not updating the frame from the focus.

This works because we know that the frames play back at a certain rate--the baseline playback rate for the animation is that 1 second of timestamp time should take 1 second to play back. So if we update the focus to match the frame with a callback that runs once every 100 ms, we should expect any discrepancy up to 100 ms between the focus and frame time to be due to the animation, so don't update the frame in response to the changed focus. (Or if we're playing at 10x speed, then a 100-ms update interval means we should expect and accept time discrepancies up to 1 sec.) (Also I multiply the window by 1.5 to give a bit of extra slush.) If the focus is updated to something outside that time window--like because the user clicked in the raster plot to change it--the difference is big enough, so it resets the position of the animation as per usual.

The upshot is that live focus updates work, in a throttled way, while still allowing clicks in the raster plot to work during playback.

The function that would nominally do this callback is complicated and most of its parameters (like the playback rate multiplier, the state dispatcher, etc.) aren't really relevant to when it should run, so to make the relationship clearer I've written the update function as a callback that takes the external time (focus time). This gets returned by the `useLiveTimeSyncing` hook as the `handleOutsideTimeUpdate` function returned in l. 115 below. That in turn gets called in a `useEffect` hook in `TrackPositionAnimationView` l. 154--which `useEffect` hook fires only when the focus time is updated.

This was all a bit complicated; I hope that this note explains more than it confuses...

jsoules:
Forgot to mention some example URLs.

See https://www.figurl.org/f?v=gs://figurl/spikesortingview-8dev&d=sha1://ff184ea76343c6fb036fdd0ac75154868317567d&label=Time%20click%20test&s={} for an example of a synced view that can show off syncing and live-focus-update features, as well as playback windowing.

See https://www.figurl.org/f?v=gs://figurl/spikesortingview-8dev&d=sha1://464b1db0ebc23408c816126c3ff67b6e1d21a77d&label=chimi_6dev4-compatible&s={} for a still-working (in fact I had to re-export it) version of data that includes decoded position.

src/FigurlCanvas/BaseCanvas.tsx
```diff
@@ -14,7 +14,7 @@ const getDrawingContextFromCanvasRef = (canvasRef: React.MutableRefObject<HTMLCa
     return ctxt
 }
 
-type DrawFn<T> = (ctxt: CanvasRenderingContext2D, data: T) => void
+export type DrawFn<T> = (ctxt: CanvasRenderingContext2D, data: T) => void
```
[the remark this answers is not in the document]
magland: sounds good to me.

PR #67, opened 2022-08-23
Implements bookmarking of animation crop window and current frame to URL.

jsoules:
This PR implements URL interaction on the pattern of the curation widgets to add the current playback window start and end frame, and the current animation frame, to the URL. This allows sharing specific time ranges or timepoints of interest within the animation.

This can be viewed using the `gs://figurl/spikesortingview-8deva` code delivery package, e.g. https://www.figurl.org/f?v=gs://figurl/spikesortingview-8deva&d=sha1://ff184ea76343c6fb036fdd0ac75154868317567d&label=Time%20click%20test&s={%22playbackWindowFrameStart%22:57594,%22playbackWindowFrameEnd%22:74258,%22playbackWindowFocusFrame%22:66704}

Idiosyncracies:

- It's currently not possible to bookmark a window comprising the entire animation length. If this is needed, we can revisit.
- I've seen occasional behavior where loading the bookmark stored in the URL results in an infinite call loop. This is always resolved by just refreshing the page--I think it's just something weird that happens when the front-end code does a save-driven live refresh while running in dev mode on localhost--so I'm not going to worry about it too hard unless we see it in the wild.

Implementation:

- A `PlaybackBookmarkButton` has been implemented as an optional button (included on the right-hand side of the playback/scrub bar).
- Pulled the designation of optional buttons out of `FrameAnimation` (the base/default class for animations, intended to be generic across playback contexts) and made it a responsibility of the implementer (e.g. `TrackPositionAnimation` which is specific to spike sorting/these experiments). This probably should've been like this from the beginning, but it is also the neatest way to make sure that the `FrameAnimation` et al don't have a direct dependency on the `FigURL`-specific way of interacting with the URL & making sure we can support other forms of bookmarking in the future.
- Added a `useUrlPlaybackWindowSupport` hook which offers three URL-state management callbacks. One updates the current `AnimationState` to match what's described in the URL (the consumer should call this every time the state's frame data is updated--this deals with the situation we attempt to set an impossible playback state because the frame data hasn't been loaded yet). The second callback takes the current state and writes it into the URL, and the third callback takes an `AnimationState` and returns true/false based on whether it matches the current URL state. Taken altogether, these three hooks encapsulate interaction with the URL into one class, and allow the rest of the UI to interact with an abstract idea of a bookmarked or saved state rather than having a tight coupling to the URL implementation specifically. (For instance, the same interface would allow implementation of a library of different bookmarked states in local or web-based storage.)
- New dependency: Added the "regular" font-awesome set in addition to the "solid" set, since I needed the empty bookmark icon. I don't think this impacts our delivery size too much, but we can revisit if it's actually folding in the full character set.
- Updated old dependencies: Needed to run a `yarn upgrade` to make the FontAwesome changes actually stick.

### DIFF
--- a/src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
+++ b/src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
@@ -3,10 +3,15 @@ import { TwoDTransformProps, use2DTransformationMatrix, useAspectTrimming } from
 import { Margins } from 'FigurlCanvas/Geometry'
 import { matrix, Matrix, multiply, transpose } from 'mathjs'
 import React, { FunctionComponent, useCallback, useEffect, useMemo } from "react"
+import { BOOKMARK_BUTTON } from 'views/common/Animation/AnimationControls/PlaybackBookmarkButton'
+import { CROP_BUTTON } from 'views/common/Animation/AnimationControls/PlaybackCropWindowButton'
+import { PlaybackOptionalButtons } from 'views/common/Animation/AnimationControls/PlaybackOptionalButtons'
+import { SYNC_BUTTON } from 'views/common/Animation/AnimationControls/PlaybackSyncWindowButton'
+import useUrlPlaybackWindowSupport from 'views/common/Animation/AnimationControls/useUrlPlaybackWindowSupport'
 import AnimationStateReducer, { AnimationState, AnimationStateAction, makeDefaultState } from 'views/common/Animation/AnimationStateReducer'
 import useLiveTimeSyncing from 'views/common/Animation/AnimationUtilities/useTimeSyncing'
 import useTimeWindowSyncing from 'views/common/Animation/AnimationUtilities/useTimeWindowSyncing'
-import FrameAnimation from 'views/common/Animation/FrameAnimation'
+import FrameAnimation, { AnimationOptionalFeatures } from 'views/common/Animation/FrameAnimation'
 import TPADecodedPositionLayer, { useConfiguredPositionDrawFunction } from './TPADecodedPositionLayer'
 import { getDecodedPositionFramePx, useProbabilityFrames, useProbabilityLocationsMap } from './TPADecodedPositionLogic'
 import TPAPositionLayer from './TPAPositionLayer'
@@ -137,8 +142,9 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
     }, [data.decodedData])
     const dataFrames = useFrames(data.positions, decodedData, transform, headDirection, data.timestampStart, data.timestamps)
     const decodedLocationsMap = useProbabilityLocationsMap(transform, decodedData)
-    useRecordingSelectionTimeInitialization(data.timestamps[0] + (data.timestampStart ?? 0), data.timestamps[data.timestamps.length - 1] + (data.timestampStart ?? 0))
 
+    useRecordingSelectionTimeInitialization(data.timestamps[0] + (data.timestampStart ?? 0), data.timestamps[data.timestamps.length - 1] + (data.timestampStart ?? 0))
+    
     useEffect(() => {
         animationStateDispatch({
             type: 'UPDATE_FRAME_DATA',
@@ -155,6 +161,15 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
     useEffect(() => handleFrameTimeUpdate(), [handleFrameTimeUpdate])
 
     useTimeWindowSyncing(animationState, animationStateDispatch, getTimeFromFrame)
+
+    const {setStateToInitialUrl, handleSaveWindowToUrl, compareStateToUrl} = useUrlPlaybackWindowSupport(animationStateDispatch)
+    useEffect(() => setStateToInitialUrl(), [setStateToInitialUrl, dataFrames]) // triggers on (unchanging) callback or when the data is updated
+    const optionalPlaybackControls: AnimationOptionalFeatures = {
+        optionalButtons: [ SYNC_BUTTON, CROP_BUTTON, BOOKMARK_BUTTON ] as PlaybackOptionalButtons[],
+        doBookmarkCallback: handleSaveWindowToUrl,
+        checkBookmarkedCallback: compareStateToUrl
+    }
+    
 
     const currentProbabilityFrame = useMemo(() => {
         const linearFrame = animationState.frameData[animationState.currentFrameIndex]?.decodedPositionFrame
@@ -199,6 +214,7 @@ const TrackPositionAnimationView: FunctionComponent<TrackPositionAnimationProps>
             state={animationState}
             dispatch={animationStateDispatch}
             dataSeriesFrameRateHz={samplingFrequencyHz}
+            options={optionalPlaybackControls}
         >
             {trackLayer}
             {probabilityLayer}
--- a/src/views/common/Animation/AnimationControls/PlaybackBookmarkButton.tsx
+++ b/src/views/common/Animation/AnimationControls/PlaybackBookmarkButton.tsx
@@ -1,0 +1,38 @@
+import { faBookmark as hollow } from '@fortawesome/free-regular-svg-icons'
+import { faBookmark as solid } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import React, { useCallback, useMemo } from 'react'
+
+
+export const BOOKMARK_BUTTON = "bookmarkButton"
+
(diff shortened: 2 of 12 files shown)


src/views/TrackPositionAnimation/TrackPositionAnimationView.tsx
```diff
@@ -3,10 +3,15 @@ import { TwoDTransformProps, use2DTransformationMatrix, useAspectTrimming } from
 import { Margins } from 'FigurlCanvas/Geometry'
 import { matrix, Matrix, multiply, transpose } from 'mathjs'
 import React, { FunctionComponent, useCallback, useEffect, useMemo } from "react"
+import { BOOKMARK_BUTTON } from 'views/common/Animation/AnimationControls/PlaybackBookmarkButton'
+import { CROP_BUTTON } from 'views/common/Animation/AnimationControls/PlaybackCropWindowButton'
+import { PlaybackOptionalButtons } from 'views/common/Animation/AnimationControls/PlaybackOptionalButtons'
+import { SYNC_BUTTON } from 'views/common/Animation/AnimationControls/PlaybackSyncWindowButton'
```
jsoules, line 9: Optional configuration is now in the implementing class, rather than being hard-coded in the base `FrameAnimation` class (at which point it isn't actually optional).

src/views/common/Animation/AnimationControls/PlaybackBookmarkButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { faBookmark as solid } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import React, { useCallback, useMemo } from 'react'
+
+
+export const BOOKMARK_BUTTON = "bookmarkButton"
+
+type PlaybackSyncButtonProps = {
+    handleSave: () => void
+    stateIsSaved: boolean
```
jsoules, line 11: This component needs only a (parameterless) callback to execute when the user wants to bookmark the current state, and a boolean value identifying whether the current state is a bookmark (in which case we show a different icon and ignore requests to bookmark again). There isn't a way to delete bookmarks under the current implementation, but it would be a minor addition.

src/views/common/Animation/AnimationControls/PlaybackOptionalButtons.ts
```diff
@@ -1,8 +1,9 @@
 
+import { BOOKMARK_BUTTON } from './PlaybackBookmarkButton'
 import { CROP_BUTTON } from './PlaybackCropWindowButton'
 import { SYNC_BUTTON } from './PlaybackSyncWindowButton'
 
-export type PlaybackOptionalButtons = typeof SYNC_BUTTON | typeof CROP_BUTTON
+export type PlaybackOptionalButtons = typeof SYNC_BUTTON | typeof CROP_BUTTON | typeof BOOKMARK_BUTTON
```
jsoules, line 7: Registering the new optional button with the infrastructure that manages such things.

src/views/common/Animation/AnimationControls/useUrlPlaybackWindowSupport.ts
```diff
@@ -0,0 +1,69 @@
+import { useUrlState } from 'figurl/UrlStateContext'
+import React, { useCallback } from 'react'
+import { AnimationState, AnimationStateAction } from 'views/common/Animation/AnimationStateReducer'
+
+const useUrlPlaybackWindowSupport = (dispatch: React.Dispatch<AnimationStateAction<any>>) => {
```
jsoules, line 5: The goal here is to encapsulate interactions with the URL (an implementation detail that isn't the only way to handle the idea of bookmarking) and provide an abstract interface to the UI (which is why the bookmark button component only worries about 'bookmark this' or 'is this bookmarked' rather than referencing the URL directly).

We might need to flesh this interface out a little bit if we wanted to support removing the bookmarked state from the URL entirely (not something we generally do in the current model of URL query parameters) or if a different implementation needed to handle multiple possible bookmarks (e.g. local storage database).

src/views/common/Animation/AnimationControls/useUrlPlaybackWindowSupport.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+        if (playbackWindowFocusFrame) {
+            dispatch({
+                type: 'SET_CURRENT_FRAME',
+                newIndex: playbackWindowFocusFrame
+            })
+        }
+    }, [dispatch, initialUrlState])
+
+    const handleSaveWindowToUrl = useCallback((state: AnimationState<any>) => {
```
jsoules, line 43: These are callbacks that take the `AnimationState` as a parameter. So we encapsulate the URL state here (centralizing the semantics of what we call those fields) and avoid having other files take dependency on the `useUrlState()` hook. Also, since the URL changes much less often than the animation state, better to have this hook return some long-lived callbacks that don't have to be refreshed when animation state changes.

src/views/common/Animation/AnimationControls/useUrlPlaybackWindowSupport.ts
```diff
@@ -0,0 +1,69 @@
...
+    const compareStateToUrl = useCallback((state: AnimationState<any>) => {
+        const urlFrame: number | undefined = urlState.playbackWindowFocusFrame
+        const urlWindowStart: number | undefined = urlState.playbackWindowFrameStart
+        const urlWindowEnd: number | undefined = urlState.playbackWindowFrameEnd
+
+        return (urlFrame !== undefined || urlWindowStart !== undefined || urlWindowEnd !== undefined)
+            && (urlFrame === undefined || urlFrame === state.currentFrameIndex)
+            && (urlWindowStart === undefined || urlWindowStart === state.window[0])
+            && (urlWindowEnd === undefined || urlWindowEnd === state.window[1])
+    }, [urlState])
```
jsoules, line 64: Again, callback curries the URL state while leaving the `AnimationState` as a parameter, since the former updates rarely and the latter updates all the time.

src/views/common/Animation/FrameAnimation.tsx
```diff
@@ -36,8 +39,21 @@ const setupReplayRate = (realTimeReplayRateMs: number | undefined, animationStat
...
+const _handleBookmark = (callback?: (state: AnimationState<any>) => void) => {
+    const fn = callback === undefined ? () => {} : callback
+    return fn
+}
+
+
+const _checkBookmark = (callback?: (state: AnimationState<any>) => boolean) => {
+    const fn = callback === undefined ? () => false : callback
+    return fn
+}
```
jsoules, line 51: These are just a convenience, since the callbacks might not be defined (as the entire functionality is optional).

jsoules: (Not sure if the anonymous functions here are going to result in needless refreshes/rewrites, we should think about that at some point)

src/views/common/Animation/FrameAnimation.tsx
```diff
@@ -46,17 +62,23 @@ const FrameAnimation = <T, >(props: FrameAnimationProps<T>) => {
...
+    const stateIsBookmarked = useMemo(() => _checkBookmark(checkBookmarkedCallback)(state), [checkBookmarkedCallback, state])
+
     const uiFeatures = useMemo(() => {
         const currentWindowIsFullRecording = (state?.window[0] === 0 && state?.window[1] === (state?.frameData?.length - 1))
         const proposalExists = (state?.windowProposal && state?.windowProposal?.length === 2)
         const proposalMatchesWindow = (state?.windowProposal && state?.windowProposal[0] === state?.window[0] && state?.windowProposal[1] === state?.window[1])
+        // TODO: make the optional buttons list a parameter rather than hard-coded here
         return {
-            optionalButtons: [ SYNC_BUTTON, CROP_BUTTON ] as PlaybackOptionalButtons[],
+            optionalButtons: optionalButtons ?? [],
```
jsoules, line 73: Don't hard-code these :)

src/views/common/Animation/SecondaryPlaybackControls.tsx
```diff
@@ -32,6 +33,10 @@ const GetSecondaryPlaybackControls = (props: SecondaryControlButtonsProps) => {
         isCropped: (ui?.isCropped ?? false),
         willCrop: (ui?.couldCrop ?? false)
     }) // TODO: Add range for display ?
+    const bookmarkButton = PlaybackBookmarkButton({
+        handleSave: (ui?.doBookmark === undefined ? () => {} : ui.doBookmark),
```
jsoules, line 37: Probably should memoize this

magland:
@jsoules is this ready to merge in your opinion?

PR #69, opened 2022-08-23
Commenting a confusing function.

jsoules:
This may be too much explanation--it's hard to find the balance.

Let me know if there are other parts of this you feel need more exposition--I could see the range scaling stuff in lines 164-166 might also qualify--this is pretty dense.

### DIFF
--- a/src/views/common/TimeScrollView/YAxisTicks.tsx
+++ b/src/views/common/TimeScrollView/YAxisTicks.tsx
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
         .filter(x => x > min)
 }
 
+/**
+ * Find a set of appropriately-scaled grid lines for the Y axis of a plot.
+ * We assume that we have data over some range, and want to break the Y axis into ticks
+ * that will generate enough grid lines to be useful, but not so many that they will make
+ * it hard to read the plot or otherwise obscure the data.
+ * 
+ * Grid lines are considered valid if each line represents a step of 1, 2, or 5 base units,
+ * where the base unit is some power of 10 of the data displayed in the chart. Conceptually,
+ * if we have measurements from (e.g.) 123 ml to 187 ml, we will first see if 1-ml grid lines work;
+ * this is probably too many, so we'll then try having each grid line/y-axis tick indicate 2 ml;
+ * then 5 ml; then 10 ml, 20 ml, etc. until we eventually find an acceptable size.
+ * 
+ * @param minGridLines Minimum number of grid lines that should be present on the plot.
+ * @param maxGridLines Maximum number of grid lines before the plot appears too busy.
+ * @param range The range of the values displayed in the plot, normalized to ensure it's got 3 digits
+ * before the decimal point.
+ * @returns {step: number, scale: number} a dictionary indicating the scale (power of 10) of the chosen
+ * grid lines and the size of each step (1, 2, or 5) at that order of magnitude. The distance between
+ * any two grid line ticks would then be step * 10 ** scale.
+ */
 const fitGridLines = (minGridLines: number, maxGridLines: number, range: number): {step: number, scale: number} => {
     let scale = 0
     while (true) {
-        const steps = [1, 2, 5]
-        const realizedScale = Math.pow(10, scale)
+        const steps = [1, 2, 5]  // candidate step sizes are 1, 2, 5, (10, 20, 50, 100, 200, ...)
+        const realizedScale = Math.pow(10, scale)   // realizedScale = the base order of magnitude represented by `scale`.
         const results: {step: number, scale: number}[] = []
-        for (let s of steps) {
+        for (let s of steps) {  // for each of the 3 candidate step sizes at this order of magnitude
+            // determine how many steps of [1 | 2 | 5] * 10^scale would be needed to exhaust the range.
+            // (technically on the plot this will be the number of spaces between the grid lines, thus the hard </>).
             const fit = range/(s * realizedScale)
             if (fit > minGridLines && fit < maxGridLines) {
+                // if this split produces an acceptable number of counts, record the step and scale sizes that work.
                 results.push({step: s, scale: scale})
             }
-            // this means the step size is too big. This shouldn't really happen without finding an acceptable step size first,
-            // but we'll check for it later just in case.
-            if (fit < minGridLines) { return {step: -1, scale: -1} }
-            results.push({step: 0, scale: 0})
+            if (fit < minGridLines) {
+                // step size was too big--it resulted in too few divisions of the range.
+                results.push({step: -1, scale: -1})
+                break
+            }
         }
+        // Completing this loop means we finished evaluating one order of magnitude. If we found
+        // a step-scale combination that gave a number of divisions in the acceptable range,
+        // it should show up as a result record with a positive step size (...viz, 1, 2, or 5).
+        // We never have to worry about it actually being from a different scale, because we run
+        // this operation every time we complete an order of magnitude.
         const a = results.find(r => r.step > 0)
         if (a) return a
+        // On the other hand, if the result set includes an entry with a step size of -1, then
+        // the last loop's step size got too big. This is bad--if we see this, it means it
+        // is not productive to try larger step sizes, but if we had found something that did
+        // work, we would've returned that first. So realistically this probably means the
+        // range of allowed grid lines is too small. (Those are set programmatically from
+        // a range of allowable inter-tick pixel counts, minGridSpacingPx and maxGridSpacingPx,
+        // which might need to be adjusted.)
         const b = results.find(r => r.step === -1)
         if (b) return b
-        scale = scale + 1
+        scale = scale + 1 // Try the next larger order of magnitude.
+        // If a step size of 50_000_000_000 base units still produces too many lines, give up.
+        // (This should never happen, since we normalized the input range to 100-1000.)
         if (scale > 10) {
             return {step: 1, scale: 0}
         }


src/views/common/TimeScrollView/YAxisTicks.tsx
```diff
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
...
+ * where the base unit is some power of 10 of the data displayed in the chart. Conceptually,
+ * if we have measurements from (e.g.) 123 ml to 187 ml, we will first see if 1-ml grid lines work;
+ * this is probably too many, so we'll then try having each grid line/y-axis tick indicate 2 ml;
+ * then 5 ml; then 10 ml, 20 ml, etc. until we eventually find an acceptable size.
+ * 
+ * @param minGridLines Minimum number of grid lines that should be present on the plot.
+ * @param maxGridLines Maximum number of grid lines before the plot appears too busy.
+ * @param range The range of the values displayed in the plot, normalized to ensure it's got 3 digits
+ * before the decimal point.
+ * @returns {step: number, scale: number} a dictionary indicating the scale (power of 10) of the chosen
```
magland, line 51: "power of 10" -> "exponent in power of 10"

src/views/common/TimeScrollView/YAxisTicks.tsx
```diff
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
...
-        const steps = [1, 2, 5]
-        const realizedScale = Math.pow(10, scale)
+        const steps = [1, 2, 5]  // candidate step sizes are 1, 2, 5, (10, 20, 50, 100, 200, ...)
+        const realizedScale = Math.pow(10, scale)   // realizedScale = the base order of magnitude represented by `scale`.
         const results: {step: number, scale: number}[] = []
-        for (let s of steps) {
+        for (let s of steps) {  // for each of the 3 candidate step sizes at this order of magnitude
+            // determine how many steps of [1 | 2 | 5] * 10^scale would be needed to exhaust the range.
+            // (technically on the plot this will be the number of spaces between the grid lines, thus the hard </>).
             const fit = range/(s * realizedScale)
```
magland, line 64: should document what fit is. Is it "estimated number of grid lines for this step and scale"?
The reason I use "estimated" is because the actual number of gridlines depends on more than just the range.

src/views/common/TimeScrollView/YAxisTicks.tsx
```diff
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
...
             if (fit > minGridLines && fit < maxGridLines) {
```
magland, line 65: Probably should use >= and <= here.

src/views/common/TimeScrollView/YAxisTicks.tsx
```diff
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
...
+                // if this split produces an acceptable number of counts, record the step and scale sizes that work.
                 results.push({step: s, scale: scale})
             }
-            // this means the step size is too big. This shouldn't really happen without finding an acceptable step size first,
-            // but we'll check for it later just in case.
-            if (fit < minGridLines) { return {step: -1, scale: -1} }
-            results.push({step: 0, scale: 0})
+            if (fit < minGridLines) {
+                // step size was too big--it resulted in too few divisions of the range.
```
magland, line 70: Expand explanation here. Need to indicate that we're going to detect the step=-1 below. And why.

src/views/common/TimeScrollView/YAxisTicks.tsx
```diff
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
...
         if (a) return a
+        // On the other hand, if the result set includes an entry with a step size of -1, then
+        // the last loop's step size got too big. This is bad--if we see this, it means it
+        // is not productive to try larger step sizes, but if we had found something that did
+        // work, we would've returned that first. So realistically this probably means the
+        // range of allowed grid lines is too small. (Those are set programmatically from
+        // a range of allowable inter-tick pixel counts, minGridSpacingPx and maxGridSpacingPx,
+        // which might need to be adjusted.)
         const b = results.find(r => r.step === -1)
         if (b) return b
```
magland, line 90: In this case the step size was too large and we are returning `{step: -1, scale: -1}`. The docs for the return value doesn't indicate this case.
Also, don't we want to allow negative values of scale? For example, a range of 0.008 or something. I think there should be a `scale = scale - 1` to try the next smaller order of magnitude.
Then we could check for both cases `scale > 10` and `scale < -10`.

src/views/common/TimeScrollView/YAxisTicks.tsx
```diff
@@ -32,27 +32,65 @@ const range = (min: number, max: number, step: number, base: number) => {
...
-        scale = scale + 1
+        scale = scale + 1 // Try the next larger order of magnitude.
+        // If a step size of 50_000_000_000 base units still produces too many lines, give up.
+        // (This should never happen, since we normalized the input range to 100-1000.)
         if (scale > 10) {
             return {step: 1, scale: 0}
```
magland, line 95: I vote for returning undefined here. We could not find a proper scale. The calling code can then figure out what to do with this. I don't like signalling by returning `{step: 1, scale: 0}`.
Also, if we handled negative values of scale, then we would never have to return the other signal `{step: -1, scale: -1}`.

src/views/common/TimeScrollView/YAxisTicks.tsx
magland: HOWEVER, we want to avoid infinite loop at all costs. For example, we don't want to get into a increment/decrement infinite cycle. This can happen depending on the values of minGridLines and maxGridLines. If we make that criteria too narrow, then you'll get the infinite back-and-forth. We need to theorize a bit about this (or use a different mechanism).

magland (changes requested):

src/views/common/TimeScrollView/YAxisTicks.tsx
jsoules: I think I had some reason for choosing strict equality, probably related to the logic that `fit` (= range over stepsize) determines the number of inter-grid-line intervals, so you wind up having to round it off... but ultimately, I think it doesn't really make a difference because both fit and `minGridLines`/`maxGridLines` (= window `pixelHeight` over the max/min number of pixels between lines) are probably non-integral anyway, so the chances of actual equality are pretty low.

magland: won't make much of a difference

src/views/common/TimeScrollView/YAxisTicks.tsx
jsoules: This isn't signaling--it's just defaulting to a unit stepsize.

magland: ah, okay.

src/views/common/TimeScrollView/YAxisTicks.tsx
jsoules: The input data range is scaled to (100, 1000] in l. 164-166:
```Typescript
        const rangeScale = Math.round(Math.log10(dataRange))
        const zoomedRangeScale = 3 - rangeScale // make sure we're counting through the range with whole numbers
        const zoomedRange = Math.round(dataRange * (Math.pow(10, zoomedRangeScale)))
```
and whatever scale comes out of `fitGridLines` then gets corrected by taking that scaling factor back out (in line 175).

So it shouldn't ever be possible to need a stepsize smaller than 1--practically, I expect the `fitGridSize()` function should almost always be returning something in the 20-200 range, but it will depend on how tight the tolerances are for pixel heights and how squashed you make the graph.

I'll annotate this part more as well.

I think the bigger issue we were running into, practically speaking, is that the loop in `fitGridSize` *returned* the -1 value immediately instead of letting it get picked up in the `while` loop--so since we don't look for a "successful" step size until we're done with checking [1, 2, 5] at that order of magnitude, if stepsize 2 works and stepsize 5 is too big, then I was returning a failure instead of catching that 2 was acceptable.

magland: I see. I think that's confusing, but I guess it should work.
So there was a bug in the prior version?

jsoules: Yeah. I corrected it in this PR by changing it to pushing `step: -1` into the results instead of returning it right away.

magland:
The following is a simpler approach, I think, in semi-pseudo-code:

```
step = 10^( floor( log(range / maxLines) ) ) // this will probably be too low for a step, but close
candidates = [step, step *2, step * 5, step * 10]
numLines = candidates.map(c => (range / c))
return numLines.find(x => (x < maxLines)) // this is the step to use
```

jsoules:
I replaced the function in question with a fleshed-out version of Jeremy's suggestion from https://github.com/magland/spikesortingview/pull/69#issuecomment-1225988010.

I ran tests between this and the earlier version; in 40k trials with random data ranges over several orders of magnitude, the results were the same for all but 19, with those being cases where the new code prefers stepping by 2*10^x instead of 1*10^x. I didn't try a whole lot of ranges of min/max grid line counts, but otherwise I think this was a pretty fair comparison.

Since the results are so close, I'm happy to use Jeremy's shorter version.

Note that I've included a check against the `minGridLines` count, so if this code returns too large a step size, we will still complain. If this actually comes up in practice, we can rework to move that check into the `simplerGridFit` function and make sure that we return another candidate if one is available.

jsoules:
For the record, test was something like this.
With `fitGridLines` as in the earlier version of this PR, in a repl:

```Typescript

const getRangeScale = (dataRange) => {
    const scale = Math.round(Math.log10(dataRange))
    const zoomedScale = 3 - scale
    return zoomedScale
}
  
let findGridsOrig = (dataRange, minGridLines, maxGridLines) => {
    const zoomedRangeScale = getRangeScale(dataRange)
    const zoomedRange = Math.round(dataRange * (Math.pow(10, zoomedRangeScale)))
    const gridInfo = fitGridLines(minGridLines, maxGridLines, zoomedRange)
    const scaledStep = gridInfo.step * Math.pow(10, gridInfo.scale - zoomedRangeScale)
    return {step: gridInfo.step, scale: gridInfo.scale - zoomedRangeScale}
}
  
let findGridsNew
findGridsNew = (dataRange, maxGridLines) => {
    const baseStep = Math.floor( Math.log10(dataRange / maxGridLines))
    const step = Math.pow(10, baseStep)
    const candidates = [1, 2, 5, 10]
    const numLines = candidates.map(c => (dataRange / (step * c)))
    const i = numLines.findIndex(x => (x < maxGridLines))
    return i === 3 ? {step: 1, scale: baseStep + 1 } : { step: candidates[i], scale: baseStep }
}
  
minGridLines = 216 / 60
maxGridLines = 216 / 23

let fullRange

let test
test = (iterations, fullRange) => {
    for (let i = 0; i < iterations; i++) {
        const thisRange = Math.random() * fullRange
        let orig = findGridsOrig(thisRange, minGridLines, maxGridLines)
        let newv = findGridsNew(thisRange, maxGridLines)
        if (newv.step === orig.step && newv.scale === orig.scale) {
            results.push(`range: ${thisRange} matched`)
        } else {
            results.push(`range: ${thisRange} old: ${orig.step}, ${orig.scale} new: ${newv.step}, ${newv.scale}`)
        }
    }
}
  

let test2
test2 = () => {
    results = []
    for (const s of [0.00099, 99.99, 9999, 999999999]) {
        test(10000, s)
    }
}
```

then manually inspect `results` and also do
```javascript
for(let x of results) {
    if (x.includes("matched")) continue
    console.log(x)
}
```

magland:
nice test!

magland (approved):

PR #70, opened 2022-08-24
Time offset

magland:
@jsoules this is a bit tricky so I think we should chat about it in person.

My goal was to allow timeOffset to be passed into PositionPlot. The timestamps would then be relative to this offset, starting with timestamp 0.

I wanted to make minimal changes to code, and avoid floating point overflow. Thus, from the perspective of the position plot view, there is no time offset. But it is added as a parameter to two hooks: `useRecordingSelectionTimeInitialization` and `useTimeRange`. The other modifications to PositionPlotView are not real changes (just removing the `data.` prefix to the props)

### DIFF
--- a/src/contexts/RecordingSelectionContext.ts
+++ b/src/contexts/RecordingSelectionContext.ts
@@ -61,24 +61,24 @@ export const useRecordingSelection = () => {
     return c
 }
 
-export const useRecordingSelectionTimeInitialization = (start: number, end: number) => {
+export const useRecordingSelectionTimeInitialization = (start: number, end: number, timeOffset: number=0) => {
     const { recordingSelection, recordingSelectionDispatch } = useRecordingSelection()
 
     useEffect(() => {
-        if (recordingSelection.recordingStartTimeSeconds === start &&
-            recordingSelection.recordingEndTimeSeconds === end) return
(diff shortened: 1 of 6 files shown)


src/contexts/RecordingSelectionContext.ts
```diff
@@ -102,8 +102,8 @@ export const useTimeRange = () => {
         })
     }, [recordingSelectionDispatch])
     return {
-        visibleTimeStartSeconds: recordingSelection.visibleTimeStartSeconds,
-        visibleTimeEndSeconds: recordingSelection.visibleTimeEndSeconds,
+        visibleTimeStartSeconds: recordingSelection.visibleTimeStartSeconds !== undefined ? recordingSelection.visibleTimeStartSeconds - timestampOffset : undefined,
+        visibleTimeEndSeconds: recordingSelection.visibleTimeEndSeconds !== undefined ? recordingSelection.visibleTimeEndSeconds - timestampOffset : undefined,
```
jsoules, line 106: I don't follow why the timestamp offset is being subtracted from these values.

magland: This is where we should chat in person.

PR #71, opened 2022-08-24
Implement highest-probability marker dot with styling.

### LINKED ISSUE #65: Mark the highest-probability position on the decoded position animation layer.
For the decoded position animation, e.g. `TPADecodedPositionLayer.tsx` and corresponding logic.

@edeno -- a few questions to refine this.

- Are you picturing a dot like the one used for the animal position, or do you mean a special highlight for the whole track pixel?
- If it's a dot, would that go in the center of the corresponding track pixel?
- How should we resolve cases where the "highest probability" location is a tie between multiple (possibly discontinuous) track pixels?
- Do we want to suppress display of this dot if it overlaps with where the animal is? I'm thinking like, check to see if the highest-probability decoded-position identifies the track pixel that contains the observed position, or maybe some kind of radius check.

jsoules:
Fixes #65.

Changes are fairly minimal--I've updated the `TPADecodedPositionLayer` to have an additional draw function for the "peak dot" (a dot in the center of a track bin with the highest probability value) with associated styling, configured through callback (it's just using default for now). Styling includes the ability to suppress display of this peak dot, if desired. (This could be made configurable by the data, but for now the styling settings are hard-coded, like all the others.)

To support the extra data, I'm calling out where the peak dot should be placed as part of the per-frame data for a decoded position frame. To find the location, we are now sorting the incoming bins by value before converting them to pixelspace (in `TPADecodedPositionLogic.getDecodedPositionFramePx`). This seemed somewhat less hacky than looking for the peak while iterating over the bins in the draw function, and also can kind of be justified by the idea that it would allow us to combine all drawing at a given intensity level into one `context.fill()` command in the future if we wanted to.

I did a spot check and things look to be correct: compare
https://www.figurl.org/f?v=gs://figurl/spikesortingview-8&d=sha1://2fd29200ff6ac8de2a312460f6dbf6a2cc3e4b5a&label=&s={}
vs
https://www.figurl.org/f?v=gs://figurl/spikesortingview-8devb&d=sha1://2fd29200ff6ac8de2a312460f6dbf6a2cc3e4b5a&label=&s={}

### DIFF
--- a/devel/deploy_dev.sh
+++ b/devel/deploy_dev.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-TARGET=gs://figurl/spikesortingview-8dev
+TARGET=gs://figurl/spikesortingview-8devb
 
 yarn build
 gsutil -m cp -R ./build/* $TARGET/
--- a/src/views/TrackPositionAnimation/TPADecodedPositionLayer.tsx
+++ b/src/views/TrackPositionAnimation/TPADecodedPositionLayer.tsx
@@ -27,8 +27,8 @@ type PeakPositionStyling = {
 }
 
 export const defaultPeakPositionStyling: PeakPositionStyling = {
-    dotRadius: 7,
-    dotRgb: 'rgb(54, 156, 0)',
+    dotRadius: 10,
+    dotRgb: 'rgb(79, 227, 0)',
     drawPeakDot: true
 }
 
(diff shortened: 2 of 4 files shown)


jsoules:
I think this can be merged, but we should leave issue #65 open until I've also implemented the tail-line feature that Eric requested. That can be a separate PR (and will probably break this stuff out into a separate Canvas and do some other more substantive changes too)

PR #72, opened 2022-08-24
Implement confusion matrix

magland:


src/views/Autocorrelograms/AutocorrelogramsView.tsx
```diff
@@ -25,7 +26,7 @@ const AutocorrelogramsView: FunctionComponent<Props> = ({data, width, height}) =
     const [showXAxis, setShowXAxis] = useState<boolean>(false)
 
     useEffect(() => {
-        unitIdSelectionDispatch({ type: INITIALIZE_UNITS, newUnitOrder: data.autocorrelograms.map(aw => aw.unitId).sort((a, b) => idToNum(a) - idToNum(b))})
+        unitIdSelectionDispatch({ type: INITIALIZE_UNITS, newUnitOrder: sortIds(data.autocorrelograms.map(aw => aw.unitId))})
```
magland, line 29: New way of sorting unit IDs. See the new sortIds() function

src/views/ConfusionMatrix/ConfusionMatrixView.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+const defaultRange: [number, number] = [0, 1]
+
+export type HoveredInfo = {
+    unitId1: number | string
+    unitId2: number | string
+    value: number | undefined
+}
+
+const ConfusionMatrixView: FunctionComponent<Props> = ({ data, width, height }) => {
```
magland, line 22: This is the new ConfusionMatrixViwe component. It is modeled after UnitSimilarityMatrixView. It uses the same MatrixWidget

src/views/ConfusionMatrix/ConfusionMatrixViewData.ts
```diff
@@ -0,0 +1,34 @@
+import { validateObject } from "figurl"
+import { isArrayOf, isEqualTo, isNumber, isOneOf, isString } from "figurl/viewInterface/validateObject"
+
+export type ConfusionMatrixViewData = {
```
magland, line 4: The data structure for the new confusion matrix view

src/views/UnitSimilarityMatrix/MatrixWidget.tsx
```diff
@@ -3,7 +3,8 @@ import { FunctionComponent, useCallback, useMemo, useState } from "react";
 import { AffineTransform, applyAffineTransform, applyAffineTransformInv, createAffineTransform, identityAffineTransform, inverseAffineTransform, multAffineTransforms } from "./AffineTransform";
 
 type Props = {
-    unitIds: (number | string)[]
+    unitIds1: (number | string)[]
+    unitIds2: (number | string)[]
```
magland, line 7: MatrixWidget has been slightly modified to allow two different sets of units... one for the columns and one for the rows. This is necessary for the confusion matrix.

src/views/UnitsTable/UnitsTableView.tsx
```diff
@@ -169,4 +169,33 @@ const UnitsTableView: FunctionComponent<Props> = ({data, width, height}) => {
...
+                if ((!isNaN(parseInt(s1))) && (!isNaN(parseInt(s2)))) {
+                    return parseInt(s1) - parseInt(s2)
+                }
+                else return id1 < id2 ? -1 : id1 > id2 ? 1 : 0
+            }
+            else return id1 < id2 ? -1 : id1 > id2 ? 1 : 0
+        }
+    })
+}
+
```
magland, line 200: Here's the new method of sorting unitIds. It handles the 3 cases shown in the comments. The second case is relevant for confusion matrix and comparisons between sortings

jsoules: This seems like a good idea--I don't know how strong the assumptions are that we can make about what people use for unit ids, clearly it bit me earlier when I assumed they'd be numbers.

Couple things that might be worth thinking about:

- `parseInt('123A') === parseInt('123B') === 123` so if we want to handle a case where there are suffixes, we might need to change the second case a bit.
- In the third case (same leading non-numeric character) it might be worthwhile to compare the strings in a loop or something to handle multiple-letter prefixes
- I guess if they have like 'A1' and '123' those would default to string comparison? That seems reasonable--it's not clear to me what an actual sort would be for them anyway

src/views/ConfusionMatrix/ConfusionMatrixViewData.ts
```diff
@@ -0,0 +1,34 @@
...
+    type: 'ConfusionMatrix'
+    sorting1UnitIds: (number | string)[]
+    sorting2UnitIds: (number | string)[]
+    unitEventCounts: {
```
jsoules, line 8: Does this refer to units on sorting 1, sorting 2, or both? If the sortings use the same naming scheme, how do you tell them apart?

src/views/ConfusionMatrix/ConfusionMatrixViewData.ts
```diff
@@ -0,0 +1,34 @@
...
+        unitId: number | string,
+        count: number
+    }[]
+    matchingUnitEventCounts: {
+        unitId1: number | string,
+        unitId2: number | string,
+        count: number
```
jsoules, line 15: So this would be the count of the number of events that match between unit id 1 and unit id 2 (from two different sortings), and we do the full pairwise comparison between the two sortings' units?

src/views/ConfusionMatrix/ConfusionMatrixView.tsx
```diff
@@ -0,0 +1,118 @@
...
+        unitIdSelectionDispatch({ type: INITIALIZE_UNITS, newUnitOrder: allUnitIds})
+    }, [allUnitIds, unitIdSelectionDispatch])
+
+    const eventCountsForIds = useMemo(() => {
+        const ret: {[id: number | string]: number} = {}
+        for (let x of unitEventCounts) {
+            ret[x.unitId] = x.count
+        }
+        return ret
+    }, [unitEventCounts])
```
jsoules, line 47: Doesn't this potentially overwrite if the two sortings use the same naming scheme for their unit ids?

src/views/ConfusionMatrix/ConfusionMatrixViewData.ts
magland: The unit IDs are assumed to be mutually exclusive between the sortings. So, A1, A2, A3, ..., B1, B2, B3, ...

src/views/ConfusionMatrix/ConfusionMatrixViewData.ts
magland: yes

src/views/ConfusionMatrix/ConfusionMatrixView.tsx
magland: the IDs are mutually exclusive

Issue #64, opened 2022-08-22
Add keyboard support for play/pause, frame adjustment for animation view


jsoules:
Ideally, space bar would hit the play/pause button, and left/right arrows would correspond to fine frame adjustment.

PR #83, opened 2022-08-30
Decouples playback control logic from UI.

jsoules:
Previously, the state interaction for each button was coded in the same file with the component itself. This makes it difficult to provide alternate paths to trigger the effect of a button (such as by keypress).

This PR collects the button effects into a logic file and expresses them as a series of handlers that take either a `MouseEvent` or a `KeyboardEvent`. The handlers depend only on the animation state reducer, so they should update infrequently. Action logic can then be injected into the various UI components, making them lighter-weight as a result. But most importantly, the state-changing logic is now available at the level of the `FrameAnimation` component. This allows attaching the keypress handlers to the largest possible structural element (the entire animation widget) instead of to individual buttons, which would lead to problems with focus.

Addresses issue #64.

### DIFF
--- a/src/views/common/Animation/AnimationControls/PlaybackControlButtonLogic.ts
+++ b/src/views/common/Animation/AnimationControls/PlaybackControlButtonLogic.ts
@@ -1,0 +1,80 @@
+import React, { useCallback, useMemo } from "react"
+import { AnimationStateDispatcher } from "../AnimationStateReducer"
+
+export type PlaybackButtonLogicCallback = (e: React.MouseEvent<any> | React.KeyboardEvent<HTMLDivElement>) => void
+export type KeyHandler = (e: React.KeyboardEvent<HTMLDivElement>) => void
+
+export type PlaybackControlLogic = {
+    returnToBeginningHandler: PlaybackButtonLogicCallback
+    backSkipHandler: PlaybackButtonLogicCallback
+    forwardSkipHandler: PlaybackButtonLogicCallback
+    jumpToEndHandler: PlaybackButtonLogicCallback
+    playPauseHandler: PlaybackButtonLogicCallback
+    reverseDirectionHandler: PlaybackButtonLogicCallback
+    syncWindowHandler: PlaybackButtonLogicCallback
+    cropWindowHandler: PlaybackButtonLogicCallback
+    keyboardControlHandler: KeyHandler
+}
+
+const usePlaybackControlButtonLogic = (dispatch: AnimationStateDispatcher<any>): PlaybackControlLogic => {
+    const handleSkipArrow = useCallback((mode: 'end' | 'skip', backward?: boolean) => {
+        return (e: React.MouseEvent | React.KeyboardEvent) => {
+            dispatch({
+                type: mode === 'end' ? 'TO_END' : 'SKIP',
+                backward: backward
+            })
+        }
+    }, [dispatch])
+
+    const returnToBeginningHandler = useMemo(() => handleSkipArrow('end', true), [handleSkipArrow])
+    const backSkipHandler = useMemo(() => handleSkipArrow('skip', true), [handleSkipArrow])
+    const forwardSkipHandler = useMemo(() => handleSkipArrow('skip'), [handleSkipArrow])
+    const jumpToEndHandler = useMemo(() => handleSkipArrow('end'), [handleSkipArrow])
+
+    const playPauseHandler = useCallback((e: React.MouseEvent | React.KeyboardEvent) => {
+        dispatch({type: 'TOGGLE_PLAYBACK'})
+    }, [dispatch])
+
+    const reverseDirectionHandler = useCallback((e: React.MouseEvent | React.KeyboardEvent) => {
+        dispatch({type: 'REVERSE_REPLAY_RATE'})
+    }, [dispatch])
+
+    const syncWindowHandler = useCallback((e: React.MouseEvent | React.KeyboardEvent) => {
+        dispatch({type: 'TOGGLE_WINDOW_SYNC'})
+    }, [dispatch])
+
+    const cropWindowHandler = useCallback((e: React.MouseEvent | React.KeyboardEvent) => {
+        dispatch({type: 'COMMIT_WINDOW'})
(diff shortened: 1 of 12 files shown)


jsoules:
See e.g. https://www.figurl.org/f?v=http://localhost:3000&d=sha1://ff184ea76343c6fb036fdd0ac75154868317567d&label=Time%20click%20test&s={}

src/views/common/Animation/AnimationControls/PlaybackControlButtonLogic.ts
```diff
@@ -0,0 +1,80 @@
...
+    returnToBeginningHandler: PlaybackButtonLogicCallback
+    backSkipHandler: PlaybackButtonLogicCallback
+    forwardSkipHandler: PlaybackButtonLogicCallback
+    jumpToEndHandler: PlaybackButtonLogicCallback
+    playPauseHandler: PlaybackButtonLogicCallback
+    reverseDirectionHandler: PlaybackButtonLogicCallback
+    syncWindowHandler: PlaybackButtonLogicCallback
+    cropWindowHandler: PlaybackButtonLogicCallback
+    keyboardControlHandler: KeyHandler
+}
```
jsoules, line 17: This mainly serves as a manifest for the types of control logic exposed here.

It may have been desirable to bring this closer to the `AnimationStateReducer`--that's worth considering in the future.

src/views/common/Animation/AnimationControls/PlaybackControlButtonLogic.ts
```diff
@@ -0,0 +1,80 @@
...
+            case 'ArrowRight':
+                fineAdjustmentHandler(e)
+                break
+            case ' ':
+                playPauseHandler(e)
+                break
+            default:
+                return
+        }
+    }, [fineAdjustmentHandler, playPauseHandler])
```
jsoules, line 73: The keyboard event handler is defined here, rather than attached to individual buttons. This has pros and cons: it might be desirable to encapsulate everything about a button a little more (by having buttons define their own keystroke), but this prevents conflicting button definitions, and also makes it easier to have one central function to attach to the `FrameAnimation` object. Also, note that some keyboard-based interactions may not have a specific button to attach to--for instance the mouse-wheel controls for fine frame adjustment aren't a clickable button, and really operate on a different paradigm from buttons anyway.

src/views/common/Animation/AnimationControls/PlaybackControlButtonLogic.ts
```diff
@@ -0,0 +1,80 @@
...
+    const fineAdjustmentHandler = useCallback((e: React.KeyboardEvent) => {
+        // This may be ill-advised
+        if (e.key !== 'ArrowLeft' && e.key !== 'ArrowRight') return
+        dispatch({
+            type: 'SKIP',
+            backward: e.key === 'ArrowLeft',
+            fineSteps: 1,
+            frameByFrame: e.ctrlKey
+        })
+    }, [dispatch])
```
jsoules, line 59: This duplicates the logic of the mouse wheel control, but given the different paradigm between debounced mouse-wheel events and individual key presses, this is hopefully the right thing to do.

src/views/common/Animation/AnimationControls/PlaybackCropWindowButton.tsx
```diff
@@ -1,23 +1,20 @@
...
-import { AnimationStateDispatcher } from 'views/common/Animation/AnimationStateReducer'
+import { useMemo } from 'react'
+import { PlaybackButtonLogicCallback } from './PlaybackControlButtonLogic'
 
 
 export const CROP_BUTTON = "cropButton"
 
 type PlaybackCropWindowButtonProps = {
-    dispatch: AnimationStateDispatcher<any>
+    cropWindowHandler: PlaybackButtonLogicCallback
```
jsoules, line 10: Individual buttons can now depend on the specific logic they require, rather than taking the entire state dispatcher and coding the logic in the button UI code.

src/views/common/Animation/AnimationControls/PlaybackPositionArrowButtons.tsx
```diff
@@ -1,37 +1,29 @@
...
             <FontAwesomeIcon icon={faForward} />
-        </span>, [handleArrowClick])
+        </span>, [forwardSkipHandler])
 
     const endButton = useMemo(() => 
-        <span onMouseDown={handleArrowClick('end')} title="Skip to end">
+        <span onMouseDown={jumpToEndHandler} title="Skip to end">
             <FontAwesomeIcon icon={faFastForward} />
-        </span>, [handleArrowClick])
+        </span>, [jumpToEndHandler])
```
jsoules, line 26: Since these no longer rely on a shared callback in this file to define their logic, it might be better to break them up into separate files. Eh.

src/views/common/Animation/AnimationStateReducer.tsx
```diff
@@ -63,6 +65,10 @@ export type AnimationStateSetReplayRateAction = {
     newRate: number
 }
 
+export type AnimationStateReverseReplayRateAction = {
+    type: 'REVERSE_REPLAY_RATE'
+}
+
```
jsoules, line 71: This had been implemented by doing a `SET_REPLAY_RATE` with the value equal to -1 times the previous rate, but I thought it would be neater to have the reducer just take care of that.

src/views/common/ViewToolbar.tsx
```diff
@@ -39,7 +39,8 @@ const ToolbarButton: FunctionComponent<ToolbarElement> = (props: ToolbarElement)
                 key={props.elementIndex}
                 color={color}
                 style={iconButtonStyle}
-                disabled={props.disabled ? true : false}>
+                disabled={props.disabled ? true : false}
+            >
```
jsoules, line 43: Makes the indentation clearer.

magland:
Would be great to have tooltip hints for the keyboard shortcuts. For example, hover over "play" button and see [spacebar]. Also would be helpful to have a help popup dialog that describes all the different key/mouse controls. (I can show how to make a popup dialog)

PR #84, opened 2022-08-30
LiveEvaluateFunctionView

magland:
A simple widget to test local tasks with electron


jsoules:
To be clear, this refers to calling `Task`-type functions that are defined by a Python back-end, right? Not allowing evaluation of arbitrary functions sent in from the UI?

magland:
> To be clear, this refers to calling `Task`-type functions that are defined by a Python back-end, right? Not allowing evaluation of arbitrary functions sent in from the UI?

Correct. And this one is just for development/testing. Minimal example.

PR #85, opened 2022-08-30
implement live traces

magland:
Implement live traces view


magland:
Implements LiveTracesView, which is the first useful example that is designed to be run locally in an electron window. I can't include an example figURL because it's meant to be run locally.
The interesting thing about it is that it uses an in-memory object to retrieve the data chunks for the traces. This allows viewing of SI recording objects without serializing them to disk.

PR #93, opened 2022-09-07
Genericize TimeScrollView and break into more manageable/reusable chunks.

jsoules:
The intent is to put this (broadly used) base component in its own directory, as well as to make the logic into easier-to-follow, more self-contained chunks. I have also implemented a `pointProjection.ts` file which hopefully can centralize and standardize pixel space conversions--more cleanup needs to be done in this area still.


src/views/SpikeAmplitudes/SpikeAmplitudesView.tsx
```diff
@@ -150,29 +150,29 @@ const SpikeAmplitudesViewChild: FunctionComponent<ChildProps> = ({data, timeseri
...
+        dataYMax: amplitudeRange.yMax,
+        yScaleFactor: ampScaleFactor,
+    })
+
+    // TODO: Figure out how to encapsulate y-axis management into timescrollview
     const yTicks = useYAxisTicks({ datamin: amplitudeRange.yMin, datamax: amplitudeRange.yMax, userSpecifiedZoom: ampScaleFactor, pixelHeight: panelHeight })
     const yTickSet = useProjectedYAxisTicks(yTicks, pixelTransform)
 
     const panels: TimeScrollViewPanel<PanelProps>[] = useMemo(() => {
+        const pixelZero = getYAxisPixelZero(pixelTransform)
```
jsoules, line 168: This is going to be ugly no matter what we do (see the prior version l. 172) but at least we can encapsulate it.

src/views/SpikeAmplitudes/SpikeAmplitudesView.tsx
```diff
@@ -150,29 +150,29 @@ const SpikeAmplitudesViewChild: FunctionComponent<ChildProps> = ({data, timeseri
...
-    )
-
+    const pixelTransform = use2dScalingMatrix({
+        totalPixelWidth: panelWidth,
+        totalPixelHeight: panelHeight,
+        dataXMin: visibleTimeStartSeconds,
+        dataXMax: visibleTimeEndSeconds,
+        dataYMin: amplitudeRange.yMin,
+        dataYMax: amplitudeRange.yMax,
+        yScaleFactor: ampScaleFactor,
```
jsoules, line 160: Note: the `yScaleFactor` logic in `pointProjection.ts` is copied from known-working code, but we will need to do more testing on the x-axis scaling. And I don't know how thoroughly I checked how these interplay with the code to maintain the aspect ratio.

jsoules: The y-axis scaling definitely will draw data points in the margins if that's where they are supposed to lie--because that's how it has always worked--but for x-axis scaling it might be smarter to just scale the underlying visible range instead.

src/views/common/TimeScrollView/TSVPaintAxes.ts
```diff
@@ -1,50 +1,10 @@
 import { TickSet } from "views/common/TimeScrollView/YAxisTicks";
-import { TimeScrollViewPanel, TimeTick } from "./TimeScrollView";
+import { TimeTick } from "./TimeAxisTicks";
```
jsoules, line 2: Removed about half the code in this file that dealt with different layers, and specified that this is the paint logic for the axes layer (that also handles the highlight spans), with its own name, as that's by far the most complex paint logic for the TSV.

src/views/common/TimeScrollView/TimeAxisTicks.ts
```diff
@@ -0,0 +1,122 @@
+import { matrix, Matrix, multiply } from 'mathjs';
```
jsoules: This is largely just broken out unchanged from the original TimeScrollView. I did update to use the new fully-encapsulated pixel conversion.

src/views/common/TimeScrollView/TimeScrollView.tsx
```diff
@@ -0,0 +1,153 @@
+import { useTimeRange } from 'contexts/RecordingSelectionContext';
```
jsoules, line 1: This has gotten so much slimmer that git no longer realizes it's the same file.

src/views/common/TimeScrollView/TimeScrollViewInteractions/TimeScrollViewEventHandlers.ts
```diff
@@ -0,0 +1,113 @@
+import { useTimeFocus, useTimeRange } from 'contexts/RecordingSelectionContext';
```
jsoules, line 1: This file pulls out the interaction/control logic that interrupted the flow of the TimeScrollView component itself. Instead here we can focus on the interactions as pure functions, and also define a hook that sets up callbacks for each that take an appropriate mouse event. Hopefully this will make it easier to see what the interactions themselves individually do, as well as putting all the dependencies in one small place.

src/views/common/TimeScrollView/TimeScrollViewInteractions/TimeScrollViewEventHandlers.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!divRef || !divRef.current) return
+    const canvases = Array.from(divRef.current.children).filter(e => e.nodeName === 'CANVAS')
+    canvases.forEach(c => {
+        c.addEventListener('wheel', (e: Event) => {
+            if ((divRef?.current as any)['_hasFocus']) {
+                e.preventDefault()
+            }
+        })
+    })
+}
```
jsoules, line 17: This guy really could go somewhere else as we might want to reuse it, but I'm not fussed for now. (For instance it should probably be run on the animation component, since that also uses mouse wheel handlers...)

src/views/common/TimeScrollView/TimeScrollViewInteractions/divRefHandling.ts
```diff
@@ -0,0 +1,17 @@
+
```
jsoules, line 1: just providing some more semantic function names and some consistency for a few bits of code that got repeated a lot.

src/views/common/TimeScrollView/TimeScrollViewInteractions/useTimeScrollZoom.ts
```diff
@@ -0,0 +1,62 @@
+import { defaultZoomScaleFactor, ZoomDirection } from 'contexts/RecordingSelectionContext';
```
jsoules, line 1: Again, this file was meant to better encapsulate a small area of functionality. I also moved the throttling to the standard framework.

src/views/common/TimeScrollView/TimeScrollViewSpans.ts
```diff
@@ -0,0 +1,43 @@
+import { Matrix } from 'mathjs';
```
jsoules, line 1: This logic had been embedded within the TimeScrollView itself, but it'll be easier to follow if it's by itself with clearly-named API functions. This was mostly a copy-unchanged situation.

PR #103, opened 2022-09-09
Experimental selector 1

magland:
prelim timepoint selector for Alex W et al

image


src/ViewData.ts
```diff
@@ -1,36 +1,39 @@
...
-import { ConfusionMatrixViewData, isConfusionMatrixViewData } from 'views/ConfusionMatrix/ConfusionMatrixViewData'
-import { isLiveEvaluateFunctionViewData, LiveEvaluateFunctionViewData } from 'views/LiveEvaluateFunction/LiveEvaluateFunctionView'
-import { isLiveTracesViewData, LiveTracesViewData } from 'views/LiveTraces/LiveTracesViewData'
+import { isSortingSelectionViewData, SortingSelectionViewData } from 'libraries/view-sorting-selection'
+import { isSpikeAmplitudesViewData, SpikeAmplitudesViewData } from "libraries/view-spike-amplitudes"
+import { isTiledImageData, TiledImageData } from 'libraries/component-tiled-image'
+import { isTrackAnimationStaticData, TrackAnimationStaticData } from "libraries/view-track-position-animation"
+import { isMainLayoutViewData, MainLayoutViewData } from 'libraries/view-main-layout'
+import { ExperimentalSelector1ViewData } from 'libraries/view-experimental-selector-1/ExperimentalSelector1ViewData'
+import { isExperimentalSelector1ViewData } from 'libraries/view-experimental-selector-1'
```
jsoules, line 36: Shouldn't these be part of the same import? Rather than referencing a non-index file from a library?

magland: you are right. Not sure why the dependency validator didn't catch that. I'll open an issue.

PR #104, opened 2022-09-09
View-timeseries-graph

magland:
implement view-timeseries-graph


magland:
Example:
https://github.com/magland/sortingview/blob/main/examples/example_timeseries_graph.py

https://www.figurl.org/f?v=gs://figurl/spikesortingview-9&d=sha1://ef6660f3a74ff8b12ef4c5d29886b33cbaf77705&label=Timeseries%20graph%20example

PR #116, opened 2022-09-12
Fixes a missed memoization for 2d matrix transform.

jsoules:
The important part of this PR is correcting a missed memoization call in the point transforms library. Also important are some clarifying/simplifying changes and decoupling the `PositionPdfPlotWidget`.

I've left some notes for ways we could make the cache system more efficient, but that's well outside the scope of this task--if we want to pursue it, it should be a separate issue first with its own prioritization.

### DIFF
--- a/src/util/pointProjection.ts
+++ b/src/util/pointProjection.ts
@@ -142,15 +142,10 @@ export const usePointWidthsFromIntervals = (scalingMatrix: Matrix, data: number[
 }
 
 
-const applyPixelMargins = (props: ScalingProps) => {
-    const {totalPixelWidth, totalPixelHeight, pixelMargins} = props
-    // const invert = !dontInvertY
-
+const applyPixelMargins = (totalPixelWidth: number, totalPixelHeight: number, pixelMargins?: PartialMargins) => {
     const pixelXMin = pixelMargins?.left ?? 0
-    // const pixelYMin = invert ? pixelMargins?.top ?? 0 : pixelMargins?.bottom ?? 0 // TODO: confirm this
     const pixelYMin = pixelMargins?.top ?? 0
     const pixelXMax = totalPixelWidth - (pixelMargins?.right ?? 0)
-    // const pixelYMax = totalPixelHeight - (invert ? (pixelMargins?.bottom ?? 0) : (pixelMargins?.top ?? 0))
     const pixelYMax = totalPixelHeight - (pixelMargins?.bottom ?? 0)
 
     return { xMin: pixelXMin, xMax: pixelXMax, yMin: pixelYMin, yMax: pixelYMax }
@@ -249,8 +244,8 @@ const matrixFromScaleOffset = (scale: scaleOffset, xScaleFactor: number, yScaleF
 
 
 /**
- * For a given data range in 2 dimensions, computes a 2 x 3 matrix which can left-multiply an augmented
- * data vector to map the data points into pixel space.
+ * For a given data range in 2 dimensions, computes a (memoized) 2 x 3 matrix which can left-multiply
+ * an augmented data vector to map the data points into pixel space.
  * This is appropriate for use with 2D data (in x and y).
  * Graphics convention places y=0 at the top of the window with increasing y values moving lower in
  * the drawing space. This is probably not what you want, so by default the transformation matrix
@@ -285,25 +280,30 @@ const matrixFromScaleOffset = (scale: scaleOffset, xScaleFactor: number, yScaleF
  * @returns A 2d scaling matrix which maps native data points into a desired drawing space.
  */
 export const use2dScalingMatrix = (props: ScalingProps) => {
-    // NOTE scaling may make us draw in the margins!!!! Think about this
     const {positiveYGrowsDownward, preserveAspectRatio} = props
-    const invert = !positiveYGrowsDownward
+    const {totalPixelWidth, totalPixelHeight, pixelMargins} = props
     const {dataXMin, dataXMax, dataYMin, dataYMax, yScaleFactor, xScaleFactor} = props
-    if (dataXMin === undefined || dataXMax === undefined || dataYMin === undefined || dataYMax === undefined) {
-        console.warn("Attempt to compute pixel projection matrix with undefined data elements. Mapping to null.")
-        return matrix([[0, 0, 0], [0, 0, 0]])
-    }
-    if (dataXMin === dataXMax && dataYMin === dataYMax) {
-        console.warn("Attempt to compute pixel projection matrix, but data has 0 width in both dimensions. Mapping to null.")
-        return matrix([[0, 0, 0], [0, 0, 0]])
-    }
-    
-    const _pixelDims = applyPixelMargins(props)
-    const dataDims = { xMin: dataXMin, xMax: dataXMax, yMin: dataYMin, yMax: dataYMax }
-    const scale = computeScales(_pixelDims, dataDims, invert, (preserveAspectRatio ?? false))
-    
-    const xform = matrixFromScaleOffset(scale, (xScaleFactor ?? 1), ((yScaleFactor ?? 1) * (invert ? -1 : 1)) )
-    return xform
+    const xformMatrix = useMemo(() => {
+        // NOTE scaling may make us draw in the margins!!!! Think about this
+        const invert = !positiveYGrowsDownward
+        if (dataXMin === undefined || dataXMax === undefined || dataYMin === undefined || dataYMax === undefined) {
+            console.warn("Attempt to compute pixel projection matrix with undefined data elements. Mapping to null.")
+            return matrix([[0, 0, 0], [0, 0, 0]])
+        }
+        if (dataXMin === dataXMax && dataYMin === dataYMax) {
+            console.warn("Attempt to compute pixel projection matrix, but data has 0 width in both dimensions. Mapping to null.")
+            return matrix([[0, 0, 0], [0, 0, 0]])
+        }
+        
+        const _pixelDims = applyPixelMargins(totalPixelWidth, totalPixelHeight, pixelMargins)
+        const dataDims = { xMin: dataXMin, xMax: dataXMax, yMin: dataYMin, yMax: dataYMax }
+        const scale = computeScales(_pixelDims, dataDims, invert, (preserveAspectRatio ?? false))
+        
+        const xform = matrixFromScaleOffset(scale, (xScaleFactor ?? 1), ((yScaleFactor ?? 1) * (invert ? -1 : 1)) )
+        return xform
+    }, [positiveYGrowsDownward, preserveAspectRatio, dataXMin, dataXMax, dataYMin, dataYMax,
+        totalPixelWidth, totalPixelHeight, pixelMargins, yScaleFactor, xScaleFactor])
+    return xformMatrix
 }
 
 
--- a/src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
+++ b/src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
@@ -1,7 +1,6 @@
 import { Checkbox } from '@material-ui/core'
 import { useRecordingSelectionTimeInitialization, useTimeRange } from 'contexts/RecordingSelectionContext'
 import { FunctionComponent, useCallback, useMemo, useState } from 'react'
-import { convert1dDataSeries, use1dScalingMatrix } from 'util/pointProjection'
 import { TimeseriesLayoutOpts } from 'View'
 import TimeScrollView, { TimeScrollViewPanel } from 'views/common/TimeScrollView/TimeScrollView'
 import { usePanelDimensions, useTimeseriesMargins } from 'views/common/TimeScrollView/TimeScrollViewDimensions'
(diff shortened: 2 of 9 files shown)


src/util/pointProjection.ts
```diff
@@ -142,15 +142,10 @@ export const usePointWidthsFromIntervals = (scalingMatrix: Matrix, data: number[
 }
 
 
-const applyPixelMargins = (props: ScalingProps) => {
-    const {totalPixelWidth, totalPixelHeight, pixelMargins} = props
-    // const invert = !dontInvertY
-
+const applyPixelMargins = (totalPixelWidth: number, totalPixelHeight: number, pixelMargins?: PartialMargins) => {
```
jsoules, line 145: Broke out the individual elements for this, although in retrospect it may not matter since this call isn't being memoized.

src/util/pointProjection.ts
```diff
@@ -285,25 +280,30 @@ const matrixFromScaleOffset = (scale: scaleOffset, xScaleFactor: number, yScaleF
...
-        return matrix([[0, 0, 0], [0, 0, 0]])
-    }
-    
-    const _pixelDims = applyPixelMargins(props)
-    const dataDims = { xMin: dataXMin, xMax: dataXMax, yMin: dataYMin, yMax: dataYMax }
-    const scale = computeScales(_pixelDims, dataDims, invert, (preserveAspectRatio ?? false))
-    
-    const xform = matrixFromScaleOffset(scale, (xScaleFactor ?? 1), ((yScaleFactor ?? 1) * (invert ? -1 : 1)) )
-    return xform
+    const xformMatrix = useMemo(() => {
```
jsoules, line 286: Breaking out all the primitive data from `props` which the memoization call depends on--I don't want a situation where we have object-reference discrepancies breaking this memoization.

src/views/PositionPdfPlot/useFetchCache.ts
```diff
@@ -82,7 +82,10 @@ const useFetchCache = <QueryType, ReturnType>(fetchFunction: (query: QueryType)
             dispatch({type: 'clear'})
         }
     }, [fetchFunction])
-    const get = useMemo(() => ((query: QueryType) => {
+    // The `get` function depends on the state, so it updates every time a reducer operation
+    // fires. This is intended: updating the `get` is how we trigger consumers that a fetch
+    // operation has completed and there's new data available in the cache.
+    const get = useCallback((query: QueryType) => {
```
jsoules, line 88: Converted the memoized function to a `useCallback`, which is semantically the same but probably expresses what we want to do more clearly.

src/views/PositionPdfPlot/useFetchCache.ts
```diff
@@ -65,11 +65,11 @@ const fetchCacheReducer = (state: FetchCacheState, action: FetchCacheAction): Fe
     }
 }
 
-const queryHash = <QueryType>(query: QueryType) => {
+const queryHash = <QueryType extends {} | null>(query: QueryType) => {
     return objectHash(query)
 }
 
-const useFetchCache = <QueryType, ReturnType>(fetchFunction: (query: QueryType) => Promise<any>): FetchCache<QueryType, ReturnType> => {
+const useFetchCache = <QueryType extends {} | null, ReturnType>(fetchFunction: (query: QueryType) => Promise<any>): FetchCache<QueryType, ReturnType> => {
```
jsoules, line 72: type updates because the linter was complaining

src/views/common/TimeScrollView/TimeScrollViewDimensions.ts
```diff
@@ -63,8 +63,9 @@ export const usePanelDimensions = (width: number, height: number, panelCount: nu
 
 export const useFocusTimeInPixels = (timeToPixelMatrix: Matrix) => {
     const {focusTime} = useTimeFocus()
-    return useMemo(() => {
+    const pixelTime = useMemo(() => {
```
jsoules, line 66: Explicit memoization.

src/views/common/TimeScrollView/TimeScrollViewDimensions.ts
```diff
@@ -19,7 +19,7 @@ export type Margins = {
 }
 
 
-export const defaultMargins = {
+const defaultMargins: Margins = {
```
jsoules, line 22: Nobody else was actually using this.

src/views/common/TimeScrollView/TimeScrollViewInteractions/TimeScrollViewEventHandlers.ts
```diff
@@ -107,7 +107,11 @@ const useTimeScrollEventHandlers = (leftMargin: number, panelWidth: number, divR
     const handleMouseMove = useMouseMoveHandler(divRef, clickReader, startPan, setPanUpdate)
     const handleMouseLeave = useMouseLeaveHandler(divRef, clearPan)
 
-    return {handleMouseUp, handleMouseMove, handleMouseDown, handleMouseLeave}
+    const handlers = useMemo(() => {
```
jsoules, line 110: I don't know if this makes a difference, but it's nice to be explicit rather than return potentially referentially-unequal objects.

src/views/common/TimeScrollView/TimeScrollView.tsx
```diff
@@ -78,17 +78,16 @@ const TimeScrollView = <T extends {[key: string]: any}> (props: TimeScrollViewPr
...
-    const content = (
-        <div
-            style={{width: width - toolbarWidth, height, position: 'relative'}}
-            onWheel={handleWheel}
-            onMouseDown={handleMouseDown}
-            onMouseUp={handleMouseUp}
-            onMouseMove={handleMouseMove}
-            onMouseOut={handleMouseLeave}
-        >
+    const effectiveWidth = useMemo(() => width - toolbarWidth, [width, toolbarWidth])
```
jsoules, line 81: This computation is referenced repeatedly, so I broke it out into its own thing.

src/views/common/TimeScrollView/TimeScrollView.tsx
```diff
@@ -78,17 +78,16 @@ const TimeScrollView = <T extends {[key: string]: any}> (props: TimeScrollViewPr
...
+
+    const style = useMemo(() => {
+        return {width: effectiveWidth, height, position: 'relative'} as any as React.CSSProperties},
+        [effectiveWidth, height])
+
+    const axesLayer = useMemo(() => {
```
jsoules, line 87: Also individually memoized the various layer components, just in case.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -230,9 +227,23 @@ export const allocate1d = (N: number, value: number | undefined) => {
...
+ * @returns Convenience function to convert values to proportionally colored pixels.
+ */
+const getColorForValueFn = (min: number, max: number) => {
+    const theScale = 255 / (max - min)
+    return (v: number | undefined) => {
+        if (v === undefined) return [0, 0, 0, 0]
+        const proportion = (v - min) * theScale
+        const intensity = Math.max(0, Math.min(255, 3 * Math.floor(proportion)))
+        return [intensity, intensity, 60, 255]
+    }
```
jsoules, line 246: So, previously we had a function that was fed a value which was supposed to be the observed value at that pixel, scaled to its percent value from 0-1 over the range of observed values. That led to a lot of code where it was used to do the scaling.

Instead, I decided to replace it with a closure (returned here) that embeds the data range within it, which makes the code above more straightforward. We can also handle the special case of an undefined value here, rather than having to do it above.

Theoretically, we could do this in a fully vectorized way by taking the incoming array, flattening it, transposing the result, and applying a scale-and-offset transform matrix all at one go. There are two tricks with this. First, in order to use the library matrix code, we'd have to do a first pass through the incoming data to make sure we replaced any `undefined` values with 0 (or more properly, with `data_min`, since we need to subtract that out later in scaling). Second, we'd need to invert the y-axis for the entries before doing the `getColorForValue` operation, since that operation needs to return 4 elements (the RGBA channels) for each one position in the data source, and that's not going to play nice.

Between those two hassles, and the need for preprocessing, adding an augmentation row, etc. I figured it wasn't worth actually messing with the matrix math.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -177,22 +173,23 @@ const PositionPdfPlotWidget: FunctionComponent<Props> = ({fetchSegment, startTim
...
     const panels: TimeScrollViewPanel<PanelProps>[] = useMemo(() => {
         return [{
             key: `pdf`,
             label: ``,
-            props: {} as PanelProps,
+            props: {offscreenCanvas: canvas} as PanelProps,
             paint: paintPanel
         }]
-    }, [paintPanel])
+    }, [paintPanel, canvas])
```
jsoules, line 191: Making the offscreen canvas a prop makes it a little more straightforward to run the updates.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -177,22 +173,23 @@ const PositionPdfPlotWidget: FunctionComponent<Props> = ({fetchSegment, startTim
...
-        context.scale((pixelTimes[1] - pixelTimes[0]) / imageData.width, panelHeight / imageData.height)
-        context.drawImage(canvas, pixelTimes[0], 0)
-        context.restore()
-    }, [pixelTimes, panelHeight, imageData, visibleLinearPositions, showLinearPositionsOverlay, downsampleFactor])
+        // Draw scaled version of image
+        // See: https://stackoverflow.com/questions/3448347/how-to-scale-an-imagedata-in-html-canvas
+
+        // Scaling the offscreen canvas can be done when it's drawn in, which avoids having to deal with transforms and some margin issues.
+        context.clearRect(0, 0, context.canvas.width, context.canvas.height)
+        context.drawImage(canvas, 0, 0, panelWidth, panelHeight)
```
jsoules, line 181: We can apply a scaling operation to the bitmap being brought in as part of the `drawImage` call--no need to touch the (onscreen) canvas's transform.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -70,103 +71,98 @@ const PositionPdfPlotWidget: FunctionComponent<Props> = ({fetchSegment, startTim
...
+        return document.createElement('canvas')
+    }, [])
 
     const paintPanel = useCallback((context: CanvasRenderingContext2D, props: PanelProps) => {
+        const canvas = props.offscreenCanvas
+        if (canvas === undefined) return
         if (!imageData) return
-        // Draw scaled version of image
-        // See: https://stackoverflow.com/questions/3448347/how-to-scale-an-imagedata-in-html-canvas
-        const canvas = document.createElement('canvas')
```
jsoules, line 165: I wasn't able to confirm, but I believe this was adding a new canvas to the document with every paint call. I memoized it out instead.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -70,103 +71,98 @@ const PositionPdfPlotWidget: FunctionComponent<Props> = ({fetchSegment, startTim
...
-                    imageData.data[i + 0] = 0
-                    imageData.data[i + 1] = 0
-                    imageData.data[i + 2] = 0
-                    imageData.data[i + 3] = 0
-                }
-                i += 4
+        const totalTimePoints = visibleValues.length
+        const totalLinearPositionBuckets = visibleValues[0].length
+        const data = []
+        const colorForValue = getColorForValueFn(minValue, maxValue)
```
jsoules, line 126: Pushed the `if undefined` logic into this function, which makes the below code read a little more clearly.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -70,103 +71,98 @@ const PositionPdfPlotWidget: FunctionComponent<Props> = ({fetchSegment, startTim
...
-                }
-                else {
-                    imageData.data[i + 0] = 0
-                    imageData.data[i + 1] = 0
-                    imageData.data[i + 2] = 0
-                    imageData.data[i + 3] = 0
-                }
-                i += 4
+        const totalTimePoints = visibleValues.length
+        const totalLinearPositionBuckets = visibleValues[0].length
```
jsoules, line 124: I think these are the logical values that correspond to the dimensions of the incoming data.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -70,103 +71,98 @@ const PositionPdfPlotWidget: FunctionComponent<Props> = ({fetchSegment, startTim
...
-    const dataModel = usePositionPdfDataModel(fetchSegment, numTimepoints, numPositions, segmentSize, multiscaleFactor)
+    const dataModel = usePositionPdfDataModel(fetchSegment, numPositions, segmentSize)
     const [showLinearPositionsOverlay, setShowLinearPositionsOverlay] = useState<boolean>(false)
+    
+    const rangeStartSample = useMemo(() => {
+        return visibleTimeStartSeconds === undefined ? 0 : Math.max(0, Math.floor(visibleTimeStartSeconds - startTimeSec) * samplingFrequency)
+    }, [visibleTimeStartSeconds, startTimeSec, samplingFrequency])
+    const rangeEndSample = useMemo(() => {
+        return visibleTimeEndSeconds === undefined ? 0 : Math.min(numTimepoints, Math.ceil((visibleTimeEndSeconds - startTimeSec) * samplingFrequency))
+    }, [visibleTimeEndSeconds, numTimepoints, startTimeSec, samplingFrequency])
```
jsoules, line 82: I broke some of these computations out into individual memoized quantities--I don't fully trust destructuring of return results in the context of memoization, but more importantly I hope this makes it a bit more straightforward what the quantities in question refer to.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -55,13 +62,7 @@ const usePositionPdfDataModel = (fetchSegment: (q: FetchSegmentQuery) => Promise
...
         return ret
     }, [fetchSegmentCache, numPositions, segmentSize])
-    return {
-        get,
-        numTimepoints,
-        numPositions,
-        segmentSize,
-        multiscaleFactor
-    }
+    return { get }
```
jsoules, line 65: The other parts of this were actually just being passed through from the function inputs, so I reference them directly to avoid any memoization tricks.

src/views/PositionPdfPlot/PositionPdfPlotWidget.tsx
```diff
@@ -30,12 +29,20 @@ type Props = {
...
-const usePositionPdfDataModel = (fetchSegment: (q: FetchSegmentQuery) => Promise<number[][]>, numTimepoints: number, numPositions: number, segmentSize: number, multiscaleFactor: number) => {
+const usePositionPdfDataModel = (fetchSegment: (q: FetchSegmentQuery) => Promise<number[][]>, numPositions: number, segmentSize: number) => {
     const fetchSegmentCache = useFetchCache<FetchSegmentQuery, number[][]>(fetchSegment)
+    // The fetch cache cannot be memoized, because the 'get' depends on the state.
+    // Thus it is updated every time the cache updates, including asynchronously.
+    // Alternative: consider reworking this hook to A) be able to answer subset queries without
+    // triggering a new query (e.g. if we have range (60, 120), we can return range (60, 100) at
+    // the same downsample level without a new query) and B) using a callback to populate the
+    // data rather than returning a function that returns the data (so that there won't be a need
+    // to use a lookup-function update as the trigger for a rerender/data refresh).
```
jsoules, line 45: Just a few ideas of things we might want to do with the fetch cache. (It took me a bit to really get what it was doing.) This is not a high priority, but I wanted to get it down somewhere.

magland:
incorporated into other PRs. Closing

PR #117, opened 2022-09-12
Magland/issue115

### LINKED ISSUE #115: TimeseriesGraph: Remove the dotted center line at zero for the PositionPlot.
... maybe have a parameter to turn this on, but by default it should be off.

magland:
Fixes #115


src/libraries/view-test-1/Test1View.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (!num_chars) {
+        /* istanbul ignore next */
+        throw Error('randomAlphaString: num_chars needs to be a positive integer.')
+    }
+    var text = "";
+    var possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+    for (var i = 0; i < num_chars; i++)
+        text += possible.charAt(Math.floor(Math.random() * possible.length));
+    return text;
+}
```
jsoules, line 95: Pretty sure we have several different copies of this floating around in the code base, would be a good idea to merge and centralize.

src/libraries/view-timeseries-graph/TimeseriesGraphView.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const paintPanel = useCallback((context: CanvasRenderingContext2D, props: PanelProps) => {
+        // don't display dashed zero line (Eric's request)
+        // context.strokeStyle = 'black'
+        // context.setLineDash([5, 15]);
+        // context.lineWidth = 1
+        // context.beginPath()
+        // context.moveTo(0, props.pixelZero)
+        // context.lineTo(panelWidth, props.pixelZero)
+        // context.stroke()
+        // context.setLineDash([]);
```
jsoules, line 82: Might be worth making this a toggle? (Whether at UI or code level)

src/libraries/component-scene2d/Scene2d.tsx
```diff
@@ -0,0 +1,312 @@
+import { BaseCanvas, pointInRect, RectangularRegion, Vec4 } from "libraries/figurl-canvas";
+import { dragSelectReducer } from "libraries/util-drag-select";
+import React, { FunctionComponent, useCallback, useEffect, useReducer, useState } from "react";
+
+// https://figurl.org/f?v=gs://figurl/spikesortingview-9&d=sha1://309f3b0131934091301c3e75669138a2e04e9240&label=Test1
+// https://figurl.org/f?v=http://localhost:3000&d=sha1://309f3b0131934091301c3e75669138a2e04e9240&label=Test1
+
+type T = {
```
jsoules, line 8: `T` is very commonly used for type parameters, so might be a good style idea to use a different name here. Maybe `Scene2dObjectSubtype[s]`?

src/libraries/component-scene2d/Scene2d.tsx
```diff
@@ -0,0 +1,312 @@
...
+	else return s
+}
+
+const Scene2d: FunctionComponent<Props> = ({width, height, objects, onClickObject, onDragObject, onSelectObjects, onClick}) => {
+	const [dragState, dragStateDispatch] = useReducer(dragSelectReducer, {})
+	const [draggingObject, draggingObjectDispatch] = useReducer(draggingObjectReducer, {})
+	const [activeSelectRect, setActiveSelectRect] = useState<Vec4 | undefined>()
+	const [activeMouseEvent, setActiveMouseEvent] = useState<React.MouseEvent | undefined>()
+
+	const handleSelectRect = useCallback((r: Vec4, e: React.MouseEvent | undefined) => {
```
jsoules, line 107: The `handleSelectRect`, subsequent `useEffect`, and `paint` logic are all pretty hairy--not going to fuss about it right now but I think breaking them out into smaller functions would make it easier to follow. I also wonder if more of the select-rect stuff should be provided by the `useDragSelect` library code.

src/libraries/util-drag-select/dragSelectReducer.ts
```diff
@@ -1,8 +1,9 @@
-import { Vec2, Vec4 } from "FigurlCanvas/Geometry"
+import { Vec2, Vec4 } from "libraries/figurl-canvas"
 
 export type DragSelectState = {
     isActive?: boolean,  // whether we are in an active dragging state
     dragAnchor?: Vec2, // The position where dragging began (pixels)
+    dragPosition?: Vec2, // The position where dragging ends (pixels)
```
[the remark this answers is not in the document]
magland: not sure it's worth it, but I'll open an issue so we can think about it.

src/libraries/view-test-1/Test1View.tsx
magland: Opening an issue

src/libraries/view-timeseries-graph/TimeseriesGraphView.tsx
magland: I think we should wait until we have a need for that.

src/libraries/component-scene2d/Scene2d.tsx
magland: yes, I changed this in a later version (the scene2d-documentation PR)

src/libraries/component-scene2d/Scene2d.tsx
magland: looking at the documented version might help (recent PR)

Issue #90, opened 2022-09-07
y-axis ticks: need more space for tick value readouts in position plot, possibly other plots


jsoules:
They're currently getting cut off.

edeno:
Could we bump this one to high priority?

magland:
This is addressed to some degree in PR #120.
There I allocated more space (40 pixels) to the left margin.
I also removed the dependency of the left margin on timeseriesLayoutOpts in order to guarantee that all timeseries plots line up consistently. I think this is more important than minimizing the margins.

However, this is just a temporary solution. We need to figure out how to set the left margin sensibly so that it accommodates the tick labels (variable width) and is consistent between views.

magland:
@edeno I am closing this for now, but let me know if you encounter problems, and we'll create a new issue.

PR #120, opened 2022-09-13
Y-axis-ticks

magland:
No longer show min/max values
Adjusted y-axis left margin to allow more space, and not be dependent on timeseriesLayoutOpts
This is a temporary solution - will need to think more about setting margins for timeseries views.


magland:
Addresses #90

magland:
As an aside, memoized a `return { get }`

src/libraries/component-time-scroll-view/TSVPaintAxes.ts
```diff
@@ -61,23 +21,25 @@ export const paintAxes = <T extends {[key: string]: any}>(context: CanvasRenderi
...
+    // We no longer show the min/max labels
+        // Range-end labels
+        // const stringMax = datamax.toString()
+        // const printMax = stringMax.substring(0, 5).search(".") === -1 ? 5 : 6
+        // const stringMin = datamin.toString()
+        // const printMin = stringMin.substring(0, 5).search(".") === -1 ? 5 : 6
+        // context.textBaseline = 'bottom'
+        // context.fillText(stringMax.substring(0, printMax), labelRightEdge, topMargin)
+        // context.textBaseline = 'top'
+        // context.fillText(datamin.toString().substring(0, printMin), labelRightEdge, xAxisYCoordinate)
```
jsoules, line 42: This should be an option rather than hard-coded one way or the other

magland: opening an issue

src/libraries/component-time-scroll-view/TimeScrollViewDimensions.ts
```diff
@@ -0,0 +1,83 @@
...
+        //             left: 30 + yAxisLeftMargin,
+        //             right: 20,
+        //             top: 20,
+        //             bottom: hideTimeAxis ? 20 : 50
+        //         } : {
+        //             left: 20 + yAxisLeftMargin,
+        //             right: 20,
+        //             top: 10,
+        //             bottom: hideTimeAxis ? 0 : 40
+        //         }
```
jsoules, line 54: These have depended on margin because our practice has been to draw axis labels in the margin--the space within the margins should be exactly what's available to the data drawing. So when an axis or toolbar is displayed, we need to pad the margin out a bit more in order to give somewhere for the toolbar contents to be drawn. OTOH, if you're hiding the toolbar, we allocate less space for it.

magland: This is something we need to discuss. On the one hand, we want to minimize the margins. On the other hand, we want multiple timeseries plots to line up exactly. For now, I'm just solving the immediate problem. We have to think about what to do.

Issue #130, opened 2022-09-13
TimeseriesGraph: allow float64 offset (like PositionPlot)


magland:

magland:
Waiting for confirmation from @edeno that this now works.

edeno:
It does work.

PR #132, opened 2022-09-13
timeseriesgraph time offset

magland:
This is the the front-end solution to #130
This allows a timeOffset to be provided in the timeseriesgraph data


jsoules (approved):
Seems reasonable.

PR #133, opened 2022-09-14
clean up imports

### LINKED ISSUE #123: Clean up imports
Shouldn't these be part of the same import? Rather than referencing a non-index file from a library?

_Originally posted by @jsoules in https://github.com/magland/spikesortingview/pull/103#discussion_r969788067_

magland:
Simply clean up some imports
Fixes #123


jsoules (approved):
I haven't done anything to verify that this is an exhaustive set of the library imports that might need to be reformatted :) but everything listed here looks fine. To me this is more a "as we notice things" sort of thing anyway.

PR #136, opened 2022-09-14
Put visibility logic in the useTimeFocus() hook.

jsoules:
I think we're better off having the logic for whether the focus is visible reside with the hook that defines it from the context, rather than in the display logic in TimeScrollView. This just makes that change.


magland:
sounds good to me.

PR #140, opened 2022-09-14
focus time interval

### LINKED ISSUE #139: Focus time interval
In TimeScrollView, allow user to select a time interval for focus. This is part of #111.

magland:
Fixes #139 

This is pretty straightforward. Added a focusTimeIntervalSeconds to the recording selection state. Put in the appropriate handlers. Shift+click selects the time interval

https://www.figurl.org/f?v=http://localhost:3000&d=sha1://7719fce3538437c9e739692bef2f9d1cd0118b8f&label=Timeseries%20graph%20example&s={%22annotations%22:%22sha1://27df7514e64711fb21fec8bc781f10196f95bbf7%22}

image

### DIFF
--- a/src/libraries/context-recording-selection/RecordingSelectionContext.ts
+++ b/src/libraries/context-recording-selection/RecordingSelectionContext.ts
@@ -4,6 +4,7 @@ export type RecordingSelection = {
     recordingStartTimeSeconds?: number
     recordingEndTimeSeconds?: number
     focusTimeSeconds?: number
+    focusTimeIntervalSeconds?: [number, number]
     visibleTimeStartSeconds?: number
     visibleTimeEndSeconds?: number
     selectedElectrodeIds: (number | string)[]
@@ -124,19 +125,21 @@ export const useTimeFocus = () => {
             autoScrollVisibleTimeRange: o.autoScrollVisibleTimeRange
         })
     }, [recordingSelectionDispatch])
-    const setTimeFocusFraction = useCallback((fraction: number) => {
+    const setTimeFocusFraction = useCallback((fraction: number, opts: {event: React.MouseEvent}) => {
         if (fraction > 1 || fraction < 0) {
             console.warn(`Attempt to set time focus to fraction outside range 0-1 (${fraction})`)
             return
         }
-        
+    
         recordingSelectionDispatch({
             type: 'setFocusTime',
-            focusTimeSec: timeForFraction(fraction)
+            focusTimeSec: timeForFraction(fraction),
+            shiftKey: opts.event.shiftKey
         })
     }, [recordingSelectionDispatch, timeForFraction])
     return {
         focusTime: recordingSelection.focusTimeSeconds,
+        focusTimeInterval: recordingSelection.focusTimeIntervalSeconds,
         setTimeFocus,
         setTimeFocusFraction,
         timeForFraction
@@ -187,6 +190,7 @@ type ZoomRecordingSelectionAction = {
 type SetFocusTimeRecordingSelectionAction = {
     type: 'setFocusTime',
     focusTimeSec: number,
+    shiftKey?: boolean
     autoScrollVisibleTimeRange?: boolean
 }
 
@@ -325,8 +329,8 @@ const zoomTime = (state: RecordingSelection, action: ZoomRecordingSelectionActio
 }
 
 const setFocusTime = (state: RecordingSelection, action: SetFocusTimeRecordingSelectionAction): RecordingSelection => {
-    const {focusTimeSec, autoScrollVisibleTimeRange} = action
-    const newState = { ...state, focusTimeSeconds: focusTimeSec }
+    const {focusTimeSec, shiftKey, autoScrollVisibleTimeRange} = action
+    let newState: RecordingSelection = { ...state, focusTimeSeconds: focusTimeSec, focusTimeIntervalSeconds: undefined }
     if (autoScrollVisibleTimeRange) {
         if ((state.visibleTimeStartSeconds !== undefined) && (state.visibleTimeEndSeconds !== undefined)) {
             if ((focusTimeSec < state.visibleTimeStartSeconds) || (focusTimeSec > state.visibleTimeEndSeconds)) {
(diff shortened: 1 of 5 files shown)


src/libraries/context-recording-selection/RecordingSelectionContext.ts
```diff
@@ -325,8 +329,8 @@ const zoomTime = (state: RecordingSelection, action: ZoomRecordingSelectionActio
 }
 
 const setFocusTime = (state: RecordingSelection, action: SetFocusTimeRecordingSelectionAction): RecordingSelection => {
-    const {focusTimeSec, autoScrollVisibleTimeRange} = action
-    const newState = { ...state, focusTimeSeconds: focusTimeSec }
+    const {focusTimeSec, shiftKey, autoScrollVisibleTimeRange} = action
```
jsoules, line 332: Better not to mix the UI implementation detail (whether the shift key being held down) with the internal state logic (whether to set focus time or focus interval).

I think we'd be better served here by having a separate reducer action to set the interval (and maybe it takes both ends of the interval, or is overloaded so that it infers that one of them is the current focus time if it only gets one time) and let the front-end look at the shift key to choose which to call. (This would also make it easier to implement a drag-based version if we changed our minds or wanted to make it easier for the user to modify an existing interval.)

jsoules (commented):
So I'd like to talk through basically the semantics of this a bit, some of the intended functionality, and whether "focus" is the right mechanism/name.

The "time focus" concept was originally implemented in the context of doing time-scale zoom--it represented the time point that would be centered when zooming in and out on a time series. When we added the animation track, it was natural to associate the current frame in the animation with the current focus in the 2d time series plots.

I'm not sure it's equally clear what it means for a time interval to be "focused", or how other parts of the application (mainly the animation) should expect to interact with it. I don't think we would want this interval to be equivalent to the idea of (what I've been calling) the animation crop window--you'd like to be able to set those separately, as you currently can set the zoom levels separately for the time series and animation if you want...

Also, there's a clear motivation from the original focus time definition for that to be unique ("the" focus time), while I'm not sure that restriction would necessarily apply here.

Again, this is largely semantic--the PR doesn't remove the idea of a (unique, point) focus time--I just want to make sure I'm thinking about this correctly. Maybe we call the interval SelectedInterval? Maybe it's fine as its?

I'd also like to be clear on a couple business-logic-y/state-rules questions (which the reducer should probably enforce):

- Does an explicit focus time still exist if there's a focus interval?
- If yes:
  - Does the focus time have to lie within the interval?
  - What happens if you try to set it outside the interval?
- If no:
  - How should other components (mainly animation) change their behavior if there is no focus time?
  - Should the `useFocusTime` hook provide some sort of definition for components that expect to consume a single-point focus time? If so, how do we pick it?
- What happens if there's a call to set the focus time while an interval exists--does that clear the interval? Update the interval? Does the call fail? Or something else?

More on the UI side, how do you clear a focus interval that's been set?

magland:
> So I'd like to talk through basically the semantics of this a bit, some of the intended functionality, and whether "focus" is the right mechanism/name.
> 
> The "time focus" concept was originally implemented in the context of doing time-scale zoom--it represented the time point that would be centered when zooming in and out on a time series. When we added the animation track, it was natural to associate the current frame in the animation with the current focus in the 2d time series plots.

I've been thinking of focus as synonymous with selected. So, focusTimeSec = selectedTimeSec. So to be consistent I used focusTimeIntervalSec, meaning the same as selectedTimeIntervalSec. At some point we could change focus => selected throughout if we wanted to.

> 
> I'm not sure it's equally clear what it means for a time interval to be "focused", or how other parts of the application (mainly the animation) should expect to interact with it. I don't think we would want this interval to be equivalent to the idea of (what I've been calling) the animation crop window--you'd like to be able to set those separately, as you currently can set the zoom levels separately for the time series and animation if you want...

I agree, the focusTimeInterval is independent of animation crop window and also independent of the visible time range.

> 
> Also, there's a clear motivation from the original focus time definition for that to be unique ("the" focus time), while I'm not sure that restriction would necessarily apply here.

The main purpose of the focus interval is to be able to add time interval annotations. So I think it's okay if it's unique.

> 
> Again, this is largely semantic--the PR doesn't remove the idea of a (unique, point) focus time--I just want to make sure I'm thinking about this correctly. Maybe we call the interval SelectedInterval? Maybe it's fine as its?
> 
> I'd also like to be clear on a couple business-logic-y/state-rules questions (which the reducer should probably enforce):
> 
> * Does an explicit focus time still exist if there's a focus interval?

Yes, these are largely independent

> * If yes:
>   
>   * Does the focus time have to lie within the interval?

No

>   * What happens if you try to set it outside the interval?

It's okay. However, if the user mouse-clicks to set the focus time, the behavior is that the focus interval is cleared. This is how you would clear the focus interval.


> More on the UI side, how do you clear a focus interval that's been set?

See above.

jsoules:
> I've been thinking of focus as synonymous with selected. So, focusTimeSec = selectedTimeSec. So to be consistent I used focusTimeIntervalSec, meaning the same as selectedTimeIntervalSec. At some point we could change focus => selected throughout if we wanted to.

No worries. Like I said, just want to make sure I'm thinking about it right.

> > Also, there's a clear motivation from the original focus time definition for that to be unique ("the" focus time), while I'm not sure that restriction would necessarily apply here.
> 
> The main purpose of the focus interval is to be able to add time interval annotations. So I think it's okay if it's unique.

I'm happy for it to be unique--just trying to see if we ought to plan to support multiples of them. (It's probably easier just to keep it unique, so that by itself is a good reason not to allow more than one...)

> > * Does an explicit focus time still exist if there's a focus interval?
> 
> Yes, these are largely independent

This clears up a lot of potential confusion--we intend them to be separate things that just happen to have a similar name, and they're expected to continue being separate things in the future.

> > * What happens if you try to set it outside the interval?
> 
> It's okay. However, if the user mouse-clicks to set the focus time, the behavior is that the focus interval is cleared. This is how you would clear the focus interval.

UI-wise this makes sense. I'll just explicitly echo back that from the perspective of the reducer (which is responsible for ensuring the underlying state is always valid) there is no reason to clear an interval just because the focus time was reset.

magland:
> UI-wise this makes sense. I'll just explicitly echo back that from the perspective of the reducer (which is responsible for ensuring the underlying state is always valid) there is no reason to clear an interval just because the focus time was reset.

Yes that was a sloppy implementation on my part. Doing the right way will require some effort. We're on the same page.

PR #144, opened 2022-09-15
option for display min/max on y axis

### LINKED ISSUE #124: option for display min/max on y axis
This should be an option rather than hard-coded one way or the other

_Originally posted by @jsoules in https://github.com/magland/spikesortingview/pull/120#discussion_r969828466_

magland:
Fixes #124

straightforward, exposes showYMinMaxLabels parameter in TimeScrollView


jsoules (approved):
Looks good to me

PR #148, opened 2022-09-16
packaging using branches

magland:
An approach for packaging using git branches.

I'll describe in detail.


magland:
@jsoules i'm closing this because, based on our conversation, we're going in a different direction.

PR #149, opened 2022-09-16
packaging up libraries

magland:
@jsoules as discussed, this is the first phase of packing up the "libraries" folder. For now, there are three library repos:

https://github.com/magland/figurl-interface

https://github.com/magland/figurl-core-utils

https://github.com/magland/figurl-core-components

A subset of the library folders have gone into these.

The only real changes with this PR in this repo are:
* added dependencies to the above packages in package.json
* deleted all the packaged libraries
* changed the imports throughout

After the cleanup, the dependency graph for the remaining graph looks a lot cleaner, and is dominated by the view-* modules:

image

Next step would be to create
* figurl-spike-sorting-views
* figurl-generic-views

Also need to think about the contexts. That may be tricky with versioning. It's important that the context providers (types) are stable

For now, versions of the dependency libraries are hard-coded (by virtue of referring to a particular tag). I suppose the advantage of making these proper npm packages is that we could take advantage of semantic versioning. So we may want to switch over to that in the future. For now I like the simplicity of tagging.


magland:
@jsoules one other note... with this new system, you'll probably want to get in the habit of `yarn install` after pulling down new changes, in case the dependent library versions have changed.

magland:
I'm closing this PR so I can open a new one and give a fresh summary of the changes.

PR #150, opened 2022-09-20
Packaging up libraries

magland:
Many of the library folders have been moved into the following npm packages.

https://github.com/magland/figurl-core-utils - shared utilities used by all the packages
https://github.com/magland/figurl-interface - communication between the figurl embedded visualization and the parent figurl.org window
https://github.com/magland/figurl-core-views - shared components and views that are not domain-specific
https://github.com/magland/figurl-timeseries-views - views that relate to timeseries and multi-channel signals (not necessarily ephys). Contains AnnotationContext and RecordingSelectionContext.
https://github.com/magland/figurl-spike-sorting-views - views that are specific to spike sorting. Contains SortingSelectionContext and SortingCurationContext.

Some library folders were not (yet) moved for one reason or another. The fate of those is uncertain. See: https://github.com/magland/spikesortingview/tree/packaging-up-libraries/src/libraries

The packages are published to https://www.npmjs.com/. When one of the packages is improved, the version is incremented. Assuming it is a non-breaking version bump, you can incorporate the latest version in this package by doing:

```
# replace @figurl/spike-sorting-views by the package you want to upgrade
yarn upgrade @figurl/spike-sorting-views
```

The dependency system is mysterious in some ways, so if things aren't working properly, try deleting `node_modules` and `yarn.lock` and rerun `yarn install`.

Each package of views has a test-gui app that can be used to test, develop, and debug views with hot-module-replacement within that package (without needing to push a bunch of versions to npmjs). Here is an example: https://github.com/magland/figurl-spike-sorting-views/tree/master/test-gui

When developing one of the view packages, you can start the development server for the test-gui in that package and use figurl links with `v=http://localhost:3000` as usual!

Sometimes the view you want to test is embedded in a layout that includes views from more than one package. That works too! That's because the test-gui apps import the views from the other packages as well.


magland:
The challenge now is that we have 6 repos, and growing. So we'll be submitting PRs to all of them. I suggest, though, that we manage issues on the spikesortingview "consumer project".
Development will be slowed down to some extent in managing the various packages, but hopefully this structure will pay off in the long run.
I put a lot of effort into making the hot-module-replacement work for the individual view packages, with minimal fuss. You can basically use the same development/debug cycle with the test-gui app within each one. We could even deploy the test-gui apps to the google bucket for sharing dev bundles, without incorporating changes back into spikesortingview.

magland:
I moved the track position animation to here:
https://github.com/magland/figurl-franklab-views

I did not move the position pdf plot, because I think we should wait and see about the new decode1d view

PR #156, opened 2022-10-05
Add yarn upgrade step to deploy script. Updated yarn.lock.

jsoules:
I had already rerun yarn upgrade, so that generated a new yarn.lock. Not sure if this is correct--or if it should be gitignored?

I think the change to the deployment script is what we want, though...?


magland:
The internet says we should **not** .gitignore yarn.lock, interestingly.

Merging.

